Issue #22, opened 2014-01-21
Support forking jobs


joevandyk:
I have some jobs that process a lot of data. Using a forking worker model means that each process can start with a clean slate. If the process balloons to 1GB, that data is automatically cleaned up at the end of the job.

PR #30, opened 2014-03-07
Added Que.after_fork and a rake task for forking workers

joevandyk:
Still a work in progress. No tests yet. Not sure if all this code should go in a rake task.  Probably need to modify the logging.

The `after_fork` in my test app looks like:

```
Que.after_fork do
  DB.disconnect
  Que.connection = DB
end
```

### DIFF
--- a/lib/que/rake_tasks.rb
+++ b/lib/que/rake_tasks.rb
@@ -5,6 +5,7 @@ namespace :que do
     Que.logger = Logger.new(STDOUT)
     Que.logger.level  = Logger.const_get((ENV['QUE_LOG_LEVEL'] || 'INFO').upcase)
     worker_count  = (ENV['QUE_WORKER_COUNT'] || 1).to_i
+    queue         = ENV['QUE_QUEUE'] || ''
 
     # Preload MultiJson's code for finding the most efficient json loader
     # so we don't need to do this inside each worker process.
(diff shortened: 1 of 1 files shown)


hlascelles:
Hi @chanks ... We're in need of forking Que workers, and this looks like it fits the bill. I found the commit https://github.com/chanks/que/commit/703e5b68e3b73ddf10c0ab6d28ede637113ce1ba where you removed it for lack us use. Were there any actual issues with it (lack of tests I understand was one)?

@joevandyk You migrated off it in production I understand. Can you tell us if there were any reasons?

Thanks both - we'll be looking at reimplementing this ourselves soon...

chanks:
Well, it depends on what you mean by forking workers. If you want to start a process, fork it a few times, and then start working jobs in the children, that's not really an issue, people do that already.

If you want to be continually forking the process so you can run each job in its own fork, though, you're going to run into problems. The issue was that if the parent process has any PG connections going when it forks, the child processes hang onto them. Then, when the child process exits, they close down those connections, which disrupts the parent. I'm not sure that there's a way around it, it seems to just be how Ruby works. See #22 for more discussion.

hlascelles:
Yes, this would be to run each job in its own fork (_a la Resque_). It goes a long way to negating most memory leaks (current or future) in our, or other gems', code.

Much like Resque, an `after_fork` hook to reconnect to the DB seems to work fine.

I'll work with the code from the past commit, and see what happens. Thanks!

PR #53, opened 2014-08-21
Added failing test case when `Que.mode = :sync' and using specific `run_at'

thetron:
Hey there!

I'm really liking Que thus far - and have been playing around with the ActiveJob integration, but I _think_ I might've come across a bug when `Que.mode = :sync`.

I noticed that when the `Que.connection = nil` and `Que.mode = :sync` and you attempt to enqueue a job with a specific `run_at`, Que with throw a runtime error with the message "Que connection not established!". I added a failing test case in this commit to demonstrate.

I wanted to patch it too - but i'm not 100% sure where to start digging, or if it was even a bug (and not just me :smile:). I stumbled across this whilst adding support for `enqueue_at` in ActiveJob.

### DIFF
--- a/spec/unit/pool_spec.rb
+++ b/spec/unit/pool_spec.rb
@@ -75,6 +75,14 @@ describe "Managing the Worker pool" do
         ArgsJob.enqueue(5, :testing => "synchronous", :run_at => Time.now + 60)
         DB[:que_jobs].select_map(:job_class).should == ["ArgsJob"]
       end
+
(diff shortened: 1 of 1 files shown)


LukeClancy:
I wrapped the problem code in "if not Rails.env.test?", dirty but works

@run_at:
`
@run_at = proc { DateTime.tomorrow.beginning_of_day + 14.hours + 17.minutes + 22.seconds } if not Rails.env.test?
`
and my re-enqueues in 
`
ProductUnlockJob.enqueue(true) if repeat and not Rails.env.test?
`

PR #74, opened 2015-01-12
use now() instead of 'now'

bgentry:
The argument `'now'` is evaluated when the statement is prepared, not when
it is executed. So the timestamp in the `insert_job` query will be stale
after the first execution.

Switching it to `now()` ensures that it is always evaluated when the query
is executed.

See https://github.com/bgentry/que-go/issues/1 for more details.

Not sure if you think this needs a test or not, but you'd probably have a better idea of how you want that to look if so.

P.S. I made a golang port of Que :) https://github.com/bgentry/que-go

### DIFF
--- a/lib/que/sql.rb
+++ b/lib/que/sql.rb
@@ -60,7 +60,7 @@ module Que
       INSERT INTO que_jobs
       (queue, priority, run_at, job_class, args)
       VALUES
-      (coalesce($1, '')::text, coalesce($2, 100)::smallint, coalesce($3, 'now')::timestamptz, $4::text, coalesce($5, '[]')::json)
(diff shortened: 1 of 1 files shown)


bgentry:
Not sure what this failed test is about but it doesn't look related: https://travis-ci.org/chanks/que/jobs/46681646

chanks:
I saw! Thanks for the blog post, it was very complementary.

I actually noticed this issue when I was looking at your port the other day, and tried to reproduce the 'now'/now() issue with the Que gem and couldn't do it. Regardless, I have no objection to using now().

And yeah, there are occasional test failures on Travis that I've never been able to reproduce on my machine, and that's one of them. I wouldn't worry about it.

bgentry:
Thanks! :v:

joevandyk:
@chanks could it be that the prepared statements aren't working as expected?

chanks:
Possibly? I'll open an issue to remind myself to investigate.

PR #85, opened 2015-02-05
fix que:work task to actually wait for running jobs to finish

tycooon:
It turned out that after commit 3981bd450ad8d6ecc2672cafd8fa460554971244, `que:work` rake task actually stopped waiting for jobs to finish despite claiming to do so. I made a little fix setting `Que.worker_count` to 0, however I don't understand why setting mode to :off doesn't do that.


chanks:
Thanks! Definitely an oversight on my part. I'll release a new version momentarily. And yeah, the API needs rethinking before 1.0. The mode setting hasn't been as useful as I'd hoped.

PR #88, opened 2015-02-17
Added support for `jsonb` datatype

afrojas:
Now that 9.4 is released and the docs suggest defaulting to `jsonb` ([link](http://www.postgresql.org/docs/9.4/static/datatype-json.html)), I've added an additional migration to conditionally alter `que_jobs`' `args` column to `jsonb`.

If the version of PG doesn't support `jsonb`, the migration is essentially a noop.

When building the `:insert_job` SQL statement and while casting return types, the type of `args` is determined and set, providing backwards compatibility.

Additionally I provided a quick example for how one might query against the values contained within `args`, assuming `jsonb` is used.

### DIFF
--- a/docs/inspecting_the_queue.md
+++ b/docs/inspecting_the_queue.md
@@ -79,6 +79,10 @@ If you want to use ActiveRecord's features when querying, you can define your ow
 
 Then you can query just as you would with any other model. Since the jobs table has a composite primary key, however, you probably won't be able to update or destroy jobs this way, though.
 
+Additionally, if you're running PG >=9.4 and have Que's DB version migrated to >=4, you can query within the 'args' column easily:
+
+    QueJob.where(job_class: 'SomeJobClass').where('args @> \'[1]\'')
+
 If you're using Sequel, you can use the same technique:
 
     class QueJob < Sequel::Model
--- a/lib/generators/que/templates/add_que.rb
+++ b/lib/generators/que/templates/add_que.rb
@@ -1,7 +1,7 @@
 class AddQue < ActiveRecord::Migration
   def self.up
     # The current version as of this migration's creation.
-    Que.migrate! :version => 3
+    Que.migrate! :version => 4
   end
 
   def self.down
--- a/lib/que.rb
+++ b/lib/que.rb
@@ -40,6 +40,10 @@ module Que
       @adapter || raise("Que connection not established!")
     end
 
+    def supports_jsonb?
+      execute("SELECT data_type FROM information_schema.columns WHERE table_name = 'que_jobs' AND column_name='args'")[0]['data_type'] == 'jsonb'
+    end
+
     def execute(*args)
       adapter.execute(*args)
     end
--- a/lib/que/adapters/base.rb
+++ b/lib/que/adapters/base.rb
@@ -110,7 +110,9 @@ module Que
       CAST_PROCS[1184] = Time.method(:parse)
 
       # JSON.
-      CAST_PROCS[114] = JSON_MODULE.method(:load)
+      CAST_PROCS[114]  = JSON_MODULE.method(:load)
+      # JSONB.
+      CAST_PROCS[3802] = JSON_MODULE.method(:load)
 
       # Boolean:
       CAST_PROCS[16] = 't'.method(:==)
--- a/lib/que/migrations.rb
+++ b/lib/que/migrations.rb
@@ -3,7 +3,7 @@ module Que
     # In order to ship a schema change, add the relevant up and down sql files
     # to the migrations directory, and bump the version both here and in the
     # add_que generator template.
-    CURRENT_VERSION = 3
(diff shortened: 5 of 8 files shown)


joevandyk:
what's the benefit of using jsonb here?

chanks:
Thanks. Unfortunately, this doesn't really handle the case where somebody on an old version of PG upgrades the Que version to 4 or higher and then upgrades PG to 9.4 later - there's not really a smooth JSONB upgrade path for them. It also potentially makes future upgrades/migrations messier, since they'd have to be written to work whether the args column is JSON or JSONB.

I'd feel a bit better about rolling an unconditional migration to JSONB into 1.0 (whenever I get around to finishing it), and just saying that Postgres 9.4 is a requirement for Que 1.x. But I don't think it's a given that JSONB is necessarily the right way to go anyway, since my own testing showed that JSONB takes up more space than JSON generally, and I don't know that there's a significant use case for generalized querying on the args hash (querying on specific fields can already be accomplished via expression indices).

I'm open to being persuaded either way, though, and am curious what other people think.

afrojas:
Agreed that it would be cleaner to simply require 9.4.

In terms of JSONB vs JSON, I know there are cases that JSONB takes up more space, however the increased flexibility JSONB provides trumps the space savings, in my opinion.

In terms of indexing specific fields within the current JSON column, the way I even went down this road is because the only thing I was storing in this particular job's args was a project_id, which has no field. I could start storing `{ id: 1 }` instead, but JSONB provides a solution that includes my use case and many others.

Additionally the flexibility of JSONB allows one to create ad-hoc queries, and not need to create, potentially, different indices for different jobs, removing them when needed, etc. In other words, it more closely follows the general dev flow of creating indices where they're needed, when they're needed, while maintaining the flexibility of querying however one wants.

My $0.02, thanks for Que (we love it) and taking the time for feedback.

chanks:
No problem, thanks for a well-organized pull request :)

I played with JSONB replacing JSON for a particular workload a couple months ago and found it to take up roughly twice the space, I believe it was? Those were fairly large documents, though (JSON strings around a thousand characters long). There may not be such a pronounced difference with the smaller documents you'd typically see in an args hash, but it's worth investigating.

For your specific problem, I think an expression index on the first value in an array (with a predicate on the job_class as well, I suppose) would still work, and would probably be much smaller and more performant than a JSONB index across every field in every job. I think you could also have an expression index on the args column as cast to JSONB - that might be worth experimenting with. Your queries would be a bit more complex, but it may be more maintainable than your own fork of Que, if we decide not to support JSONB.

afrojas:
I had the same thought about the size of JSON vs JSONB for the specific case of `args`. It's unlikely that you'd have so much data in that column to make a significant impact. But worth investigating for sure.

I had not thought about indexing args and casting to JSONB... I'll check that out for sure. I have no intention of maintaining my own fork of Que, just hoping to make the case for why JSONB is the right choice :)

In terms of the indices you mentioned (first value of array and how entire JSONB index of `args` would probably be gnarly), my argument is more that by having JSONB from the start, as a developer trying to figure out the best structure and indices for `args` (regardless of datatype) I can play around more easily with how things should be organized, and then create the appropriate indices.

In other words right now I'd have to create/delete/create/<repeat> indices while developing as I settle on my needs, which can be pretty annoying. JSONB allows me to skip that step, query with abandon as I develop, and then create the appropriate indices. Seems more dev-friendly to me.

chanks:
Just spent some time digging into this. Here's the script I used:

````SQL
create table jsonb_test (v jsonb);
create table json_test (v json);

INSERT INTO jsonb_test
  SELECT row_to_json(r)::jsonb
  FROM (
    SELECT floor(random() * 1000) AS integer,
           random() AS float
    FROM generate_series(1, 100000) s
  ) r;

INSERT INTO json_test
  SELECT v::json FROM jsonb_test;

-- GIN index on casted column works fine:
create index jsonb_index on jsonb_test using gin (v);
create index json_index on json_test using gin (CAST(v AS jsonb));

-- These queries can both use their respective indexes.
EXPLAIN ANALYZE
  SELECT * FROM jsonb_test WHERE v @> '{"integer":56}';
EXPLAIN ANALYZE
  SELECT * FROM json_test WHERE CAST(v AS jsonb) @> '{"integer":56}';

SELECT pg_size_pretty(pg_total_relation_size('json_test')); -- 16 MB
SELECT pg_size_pretty(pg_total_relation_size('jsonb_test')); -- 18 MB
````

We could probably find some structure of argument hashes where JSONB would be smaller than JSON. But since the index isn't an issue, and querying jobs based on arg hash contents isn't a very common use case, I'm inclined to not introduce this complexity to the library unless JSONB has significant space savings, which it doesn't look like it will.

I'm going to close this because I'm trying to clean up the issue tracker a bit, but if you or anyone else disagrees, chime in and I'll reopen.

PR #92, opened 2015-02-25
Update `error_handling` documentation.

barisbalic:
It seemed strange that the error handler wasn't being given the job that had failed, upon closer inspection of the code I noticed that the job is in fact passed to the handler.  This PR just updates the documentation to express that fact.


chanks:
Good point, thanks!

barisbalic:
:+1:

PR #96, opened 2015-03-18
Eager load Rails apps when running `rake que:work`

hmarr:
ActiveSupport's dependency autoloading isn't threadsafe, and Que uses multiple threads, which means that eager loading is necessary. Rails explicitly [prevents eager loading](https://github.com/rails/rails/blob/master/railties/lib/rails/application.rb#L445-L446) when the environment task is invoked, so we need to manually eager load the app here.

### DIFF
--- a/lib/que/rake_tasks.rb
+++ b/lib/que/rake_tasks.rb
@@ -3,6 +3,14 @@ namespace :que do
   task :work => :environment do
     require 'logger'
 
+    if defined?(::Rails) && Rails.respond_to?(:application)
+      # ActiveSupport's dependency autoloading isn't threadsafe, and Que uses
+      # multiple threads, which means that eager loading is necessary. Rails
+      # explicitly prevents eager loading when the environment task is invoked,
+      # so we need to manually eager load the app here.
+      Rails.application.eager_load!
+    end
+
     Que.logger        = Logger.new(STDOUT)
     Que.logger.level  = Logger.const_get((ENV['QUE_LOG_LEVEL'] || 'INFO').upcase)
     Que.worker_count  = (ENV['QUE_WORKER_COUNT'] || 4).to_i


hmarr:
Worth mentioning that calling `eager_load!` directly isn't ideal (see mperham/sidekiq#1791 for details).

A better approach would be moving away from depending on the `:environment` rake task, and loading Rails by requiring `config/environment.rb` (as [Hutch](https://github.com/gocardless/hutch/blob/master/lib/hutch/cli.rb#L69) and [Sidekiq](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L241) do). That way we could inject an initializer that sets `config.eager_load` to `true` (something like [this](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L237-L239)).

Happy to do that, but it'd require a more significant change to the rake tasks.

Thoughts?

chanks:
Hey, thanks for the patch. Yeah, that all sounds good to me - I don't use the rake task, personally, and am not super familiar with Rails' boot process, so I'm happy for the help :smiley:

hmarr:
On second thought, moving away from depending on `environment` might not be ideal. Loads of Rails-related rake tasks depend on `environment`, and they can be chained (e.g. `rake que:migrate que:work`). If the `environment` task is run by any earlier tasks, the application won't be eager loaded when the `work` task is run, which I could imagine causing some headaches!

So I reckon this PR should be good to merge for the time being (it's how [Resque handles it](https://github.com/resque/resque/blob/1-x-stable/lib/resque/tasks.rb#L62)).

Longer term, what do you think about moving this stuff to a `que` executable, ala [Sidekiq](https://github.com/mperham/sidekiq/blob/master/bin/sidekiq) / [Hutch](https://github.com/gocardless/hutch/blob/master/bin/hutch)? That would allow us to optionally load Rails apps, or any other ruby files you specified, making these tasks available to non-Rails environments, and bypassing the rake issue mentioned above.

chanks:
Merged! Will release 0.10.0 shortly.

The executable approach makes sense to me - I don't think I can spend a lot of time on it myself, but after looking at the options Hutch offers, I like the utility it offers, and I'd certainly merge a PR for it (especially if it introduced no new dependencies).

As an aside, I'm definitely in favor of anything that decouples Que from Rails, and I have a partially-completed gem (unpublished, since I never got it into a state where I'd have been comfortable supporting it) that extracts the railtie and rake tasks and adds tests for them. I don't know whether it would make sense to have the CLI as a separate gem, I expect probably not.

hmarr:
Sweet, thanks for the merge.

Will have a look at replacing the rake tasks with the executable when I get some time.

Issue #71, opened 2015-01-07
Allow listening on multiple queues


chrismcg:
I've been playing with Que and the ActiveJob wrapper in Rails 4.2, looks really useful so thanks to everyone who contributed so far.

It took me a while to realise that an apps Que workers will only listen to one queue at a time, defaulting to ''. Once I'd figured out I could set `QUE_QUEUE` environment variable I made some progress. However I ran into trouble again when using the new `deliver_later` functionality in ActionMailer. The job it creates puts mails on the `mailers` queue. I've worked around this in my test app by putting all my jobs on the `mailers` queue but it's not a great solution.

I can think of two solutions to this.

The simpler one would be to let `QUE_QUEUE` potentially be a list of queues separated by commas and tweak the `:lock_job` sql to use `WHERE queue in ...`

A more complicated (but more flexible) solution would be to somehow let the app override set_up_workers and create it's own workers listening to whatever queues were required. This may already be possible by setting `worker_count` to 0 and doing `Que::Worker.workers.push Que::Worker.new('foo')` as required. I'll test this out after I've finished writing this ticket.

Would you object to a PR allowing something like the following?

```ruby
# One worker for the mailer queue and two for the default queue
Que.worker_configuration = { 'mailers' => 1, 'default' => 2 }
```

joevandyk:
I've also ran into cases where multiple queues would be nice. 

ActionMailer's deliver_later puts stuff into a 'mailer' queue by default?

chrismcg:
Yes, though it's `mailers` plural: https://github.com/rails/rails/blob/master/actionmailer/lib/action_mailer/delivery_job.rb#L7

joevandyk:
eww, yes, that's going to confuse a ton of people.

chrismcg:
OK, so in `config/application.rb` I have

```
Que.worker_count = 0
Que.mode = :off
```

and then in my `config/puma.rb` (important the worker setup comes after the mode change :smile:)

```ruby
Que.mode = :async
Que::Worker.workers.push Que::Worker.new 'mailers'
Que::Worker.workers.push Que::Worker.new 'default'
Que::Worker.workers.push Que::Worker.new 'default'
```

Running the test app and putting something in both queues led to both queues being emptied :tada:

[...]

PR #101, opened 2015-03-24
Allow a user to configure workers for multiple queues.

JonRowe:
Currently, que supports running jobs from only one queue as named per `ENV['QUE_QUEUE']`, as per #71 Rails default implementation of `ActiveJob` uses two queues, `default` and `mailers`. 

This is (hopefully) the start of a conversation on how to better support Rails users who might (like me) naively assume `config.active_job.queue_adapter = :que` and `rake que:work` is enough to get everything working.

I know I could specify multiple `rake que:work` jobs to get the same effect, but I'd like to do this on 1 dyno on Heroku and this is a naive attempt at making that work.

### DIFF
--- a/lib/que/worker.rb
+++ b/lib/que/worker.rb
@@ -161,11 +161,20 @@ module Que
       private
 
       def set_up_workers
-        if worker_count > workers.count
-          workers.push(*(worker_count - workers.count).times.map{new(ENV['QUE_QUEUE'] || '')})
-        elsif worker_count < workers.count
-          workers.pop(workers.count - worker_count).each(&:stop).each(&:wait_until_stopped)
+        count = workers.count * queues.count
+        if worker_count > count
+          queues.each do |queue|
+            workers.push(*(worker_count - count).times.map{ new(queue) })
+          end
+        elsif worker_count < count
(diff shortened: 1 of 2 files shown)


chanks:
Hi, thanks for the PR.

I think the only lasting solution to this is a patch to the ActiveJob adapter that stops using named queues entirely. Named queues were only meant for situations where completely different codebases needed to use the same DB, and the ActiveJob adapter shouldn't be using them the way it is. My tentative plan right now is for named queues to be removed entirely in 1.0 anyway (replaced by the ability to have multiple job tables in the same database, which should be faster and more flexible).

I've opened an ActiveJob PR here:

https://github.com/rails/rails/pull/19498

joevandyk:
Sounds good about removing named queues in favor of mutliple job tables. Hopefully que-web (cc @statianzo) will be able to support that.

chanks:
Yeah, it's not built yet, so I don't know how the worker/job state queries will handle it either :) but I think it's an important thing to support.

statianzo:
As long as it can get access to the list of tables, que-web should be capable of handling it.

PR #112, opened 2015-05-15
Command Line Interface

chanks:
I'm opening this as a PR so people have the chance to try it out and comment on it before it's merged into master. It replaces the current Rails-specific methods of working jobs (`rake que:work` and autostarting when the server runs) with a command-line interface. This will hopefully fix a lot of the persistent problems people have been having with workers running during migrations, asset compilations, in the console, etc.

From the changelog:

> A command-line program has been added that can be used to work jobs in a more flexible manner than the previous rake task. Run `que -h` for more information.
>
> The `rake que:work` rake task that was specific to Rails has been removed in favor of the CLI, and the various QUE_* environment variables no longer have any effect.
>
> The worker pool will no longer start automatically in the same process when running the rails server - this behavior was too prone to breakage. If you'd like to recreate the old behavior, you can manually set `Que.mode = :async` in your app whenever conditions are appropriate (classes have loaded, a database connection has been established, and the process will not be forking).

Currently, the `que` command must be invoked with the path of the application to load before jobs can be worked, and actually, any number of filenames can be provided and they'll be required in order. For people using Rails, you'll probably want to invoke the command like `que config/environment`. I considered trying to check whether the command was being called in the context of a Rails app by checking for the presence of the config/environment.rb, but this felt hacky to me. I understand that most people using Que are probably using it with Rails, though, so I suppose I could be talked into it.

Also be aware that the docs haven't been updated to reflect these changes, so take them with a grain of salt.

Feedback and experimentation (and even PRs to merge into this PR) welcome! And thanks to GoCardless for bringing up the idea a while ago, I took some inspiration from [Hutch](https://github.com/gocardless/hutch).

### DIFF
--- a/bin/que
+++ b/bin/que
@@ -74,15 +74,13 @@ Que.mode          = :async
 stop = false
 %w(INT TERM).each { |signal| trap(signal) { stop = true } }
 
-at_exit do
-  $stdout.puts
-  $stdout.puts "Finishing Que's current jobs before exiting..."
-  Que.worker_count = 0
-  Que.mode = :off
-  $stdout.puts "Que's jobs finished, exiting..."
-end
-
 loop do
   sleep 0.01
   break if stop
 end
+
+$stdout.puts
+$stdout.puts "Finishing Que's current jobs before exiting..."
+Que.worker_count = 0
+Que.mode = :off
+$stdout.puts "Que's jobs finished, exiting..."
--- a/bin/que
+++ b/bin/que
@@ -10,7 +10,7 @@ options.log_level = 'info'
 options.wake_interval = 0.1
 
 OptionParser.new do |opts|
-  opts.banner = 'usage: que [options] (filename) ...'
+  opts.banner = 'usage: que [options] file/to/require ...'
 
   opts.on('-w', '--worker-count [COUNT]', Integer, "Set number of workers in process (default: 4)") do |worker_count|
     options.worker_count = worker_count
@@ -28,16 +28,26 @@ OptionParser.new do |opts|
     options.queue_name = queue_name
   end
 
-  opts.on('-v', '--version', "Print Que's version and exit") do
+  opts.on('-v', '--version', "Show Que version") do
     $stdout.puts "Que version #{Version}"
     exit 0
   end
 
-  opts.on('-h', '--help', "Show help text and exit") do
+  opts.on('-h', '--help', "Show help text") do
     $stdout.puts opts
     exit 0
   end
 end.parse!(ARGV)
+
+if ARGV.length.zero?
+  $stdout.puts <<-OUTPUT
+You didn't include any Ruby files to require!
+Que needs to be able to load your application before it can process jobs.
+(Hint: If you're using Rails, try `que config/environment.rb`)
+(Or use `que -h` for a list of options)
+OUTPUT
+exit 1
+end
 
 ARGV.each do |file|
   begin
@@ -70,6 +80,7 @@ stop = false
 %w(INT TERM).each { |signal| trap(signal) { stop = true } }
 
 at_exit do
+  $stdout.puts
   $stdout.puts "Finishing Que's current jobs before exiting..."
   Que.worker_count = 0
   Que.mode = :off
--- a/bin/que
+++ b/bin/que
@@ -5,9 +5,6 @@ require 'ostruct'
 require 'logger'
 
 options = OpenStruct.new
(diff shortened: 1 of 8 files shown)


bin/que
```diff
@@ -0,0 +1,93 @@
...
+(Or use `que -h` for a list of options)
+OUTPUT
+exit 1
+end
+
+ARGV.each do |file|
+  begin
+    # Need to add '.' to load path for relative requires
+    $LOAD_PATH << '.'
+    require file
```
godfat: Generally this is a bad idea. People should just pass `./file.rb` if they want to do so. Ruby removed `.` from $LOAD_PATH for security reason, it's not a good idea to restore the old behaviour.

bin/que
```diff
@@ -0,0 +1,93 @@
...
+  Que.logger.level = Logger.const_get(options.log_level.upcase)
+rescue NameError
+  $stdout.puts "Bad logging level: #{options.log_level}"
+  exit 1
+end
+
+Que.queue_name    = options.queue_name
+Que.worker_count  = options.worker_count
+Que.wake_interval = options.wake_interval
+Que.mode          = :async
```
godfat: I think we should probably use `||=` here to allow setting the values in the file we're loading. It could also count as a config file. However, if we specifically pass the argument to the CLI, then we probably want to overwrite it. So I think we should probably not set the default in CLI, but here. So that this would work:

``` ruby
Que.worker_count = options.worker_count || Que.worker_count || 4
```

This means we prefer CLI argument over config file over default.

bin/que
```diff
@@ -0,0 +1,93 @@
...
+
+stop = false
+%w(INT TERM).each { |signal| trap(signal) { stop = true } }
+
+at_exit do
+  $stdout.puts
+  $stdout.puts "Finishing Que's current jobs before exiting..."
+  Que.worker_count = 0
+  Que.mode = :off
```
godfat: I propose to have a `Que.stop` having:

``` ruby
Que.worker_count = 0
Que.mode = :off
```

bin/que
```diff
@@ -0,0 +1,93 @@
...
+  $stdout.puts "Que's jobs finished, exiting..."
+end
+
+loop do
+  sleep 0.01
+  break if stop
+end
```
godfat: I am not sure which one would be better, but since the program exits here, I think we don't have to use `at_exit` but just put those in the end of this script? This could probably reduce some complexity.

chanks:
Note that, per one of @godfat's suggestions, the script no longer adds '.' to the load path, so for the Rails example, it'd have to be invoked like `que ./config/environment.rb`.

j4zz:
> It replaces the current Rails-specific methods of working jobs (rake que:work and autostarting when the server runs) with a command-line interface

How does that tie in with avoiding a separate worker dyno on Heroku? Would we still do a CLI call during application startup for this case? Or can the CLI be used after application deployment to start workers in the same dyno? Sorry if I’m missing something.

godfat:
In a nutshell, if you're using `rake que:work` before, you use `que ./config/environment.rb` instead. If you want to avoid an extra process, then you have to put `Que.mode = :async` somewhere in your code. If you're using a forked server, you could put it in your after fork callback.

chanks:
Yeah, Que won't start by itself anymore at all, since doing so was too prone to issues. You can spin it up manually by setting the mode to :async, but it'll be up to you to figure out how to do so safely for your app. For people using forking webservers, the after_fork callbacks will probably be the best bet, but I imagine it'll be affected by when classes are loaded, when the DB connection is established, etc.

bin/que
```diff
@@ -0,0 +1,85 @@
...
+    $stdout.puts "Could not load file '#{file}'"
+  end
+end
+
+Que.logger ||= Logger.new(STDOUT)
+
+log_level = options.log_level || 'info'
+
+begin
+  Que.logger.level = Logger.const_get(log_level.upcase)
```
godfat: `log_level` could have been set as well. Maybe something like:

``` ruby
# Retain original log level and make default to Logger::INFO
log_level = (Que.logger && Que.logger.log_level) || Logger::INFO
Que.logger ||= Logger.new(STDOUT)
# personally I prefer $stdout though. Not very sure about the difference.
# I only know that $stdout could be reassigned, while STDOUT would
# always be referring the original stdout.
Que.logger.level = (options.log_level && Logger.const_get(options.log_level)) || log_level
```

PR #117, opened 2015-07-26
Support non-rails environment.

afirel:
For non-rails apps, rake tasks don't work since they require the ```environment``` task. This stubs the ```environment``` which should support both rails and non-rails apps then.

Based on https://github.com/QueueClassic/queue_classic/issues/62


chanks:
This shouldn't be necessary anymore with the new CLI. The rake task itself will be removed in the next major version anyhow, and we shouldn't be encouraging its use. Thanks, though!

PR #120, opened 2015-09-03
Use ActiveSupport's constantize

### LINKED ISSUE #115: Use ActiveSupport's constantize if it exists
```
$ rails c
Loading development environment (Rails 4.2.1)

irb(main):001:0> string = "Fulfiller::Jobs::SunFrog::SendOrders"
=> "Fulfiller::Jobs::SunFrog::SendOrders"

irb(main):002:0> string.split('::').inject(Object, &:const_get)
NameError: uninitialized constant SunFrog::SendOrders
	from (irb):2:in `const_get'
	from (irb):2:in `each'
	from (irb):2:in `inject'
	from (irb):2

irb(main):003:0> string.constantize
=> Fulfiller::Jobs::SunFrog::SendOrders
```

https://github.com/chanks/que/blob/master/lib/que/job.rb#L140 -- using const_get doesn't work perfectly with Rails's autoloading (I also have a top-level SunFrog constant).

joevandyk:
Fixes #115

### DIFF
--- a/lib/que.rb
+++ b/lib/que.rb
@@ -129,6 +129,38 @@ module Que
       @disable_prepared_statements || false
     end
 
+    # Copied from ActiveSupport's 4.2.4 version.
+    def constantize(camel_cased_word)
+      names = camel_cased_word.split('::')
+
+      # Trigger a built-in NameError exception including the ill-formed constant in the message.
+      Object.const_get(camel_cased_word) if names.empty?
+
+      # Remove the first blank element in case of '::ClassName' notation.
+      names.shift if names.size > 1 && names.first.empty?
+
+      names.inject(Object) do |constant, name|
+        if constant == Object
+          constant.const_get(name)
+        else
+          candidate = constant.const_get(name)
+          next candidate if constant.const_defined?(name, false)
+          next candidate unless Object.const_defined?(name)
+
+          # Go down the ancestors to check if it is owned directly. The check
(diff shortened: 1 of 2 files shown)


joevandyk:
I was not able to get any tests to show the bug. :/

chanks:
Hmm, that's disconcerting. I want to give it a shot myself before merging.

joevandyk:
Another option is to use Sidekiq's version: 
https://www.omniref.com/ruby/gems/sidekiq/0.6.0/symbols/Sidekiq::Util/constantize

Resque seems to copy over the implementation: 
https://github.com/resque/resque/blob/master/lib/resque/core_ext/string.rb

I updated the commit with using String#constanize if it exists. I'm not sure what the best method is.

chanks:
Tried a bit to write a failing spec and couldn't, but oh well, it's not a bad idea anyways.

Will be releasing 0.11.0 in just a bit.

joevandyk:
I wasn't sure if you would want to use the methods that Sidekiq or Resque use. (It's too bad this isn't a standard Ruby method)

chanks:
I think keeping the current method if ActiveSupport isn't present is fine.
If someone not using ActiveSupport runs into it we can import one of those
methods later.

And I just realized I forgot to update the changelog with this fix before I
cut the new version, but I'll do it now.

PR #121, opened 2015-09-09
Fix undefined string

godfat:
Seems like we have less coverage without Rails...

### DIFF
--- a/lib/que.rb
+++ b/lib/que.rb
@@ -134,7 +134,7 @@ module Que
         # Use ActiveSupport's version if it exists.
         camel_cased_word.constantize
       else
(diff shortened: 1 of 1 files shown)


chanks:
Thanks! Yes, the `rake pg` task runs the unit specs without ActiveSupport loaded, and would have caught this bug, but I let a rogue activesupport require sneak into the unit specs a little while back (it was part of a PR, but I should have caught it). Sorry about that :frowning: 

Fixed the specs, merged this, and released 0.11.2. Thanks again!

godfat:
No problems and no worries, thank you! :D

PR #122, opened 2015-09-21
Fix uninitialized constant Version (NameError).

hardbap:
Running `que --version` throws uninitialized constant Version (NameError)


chanks:
Thanks!

PR #123, opened 2015-09-21
Use wake_interval option in loop

hardbap:
Currently the loop is only sleeping 0.01 seconds and cannot be changed.


chanks:
That loop isn't for polling the DB (which is what wake_interval is used for), it's for checking for whether it's time to end the script. Unless having it as a static value is giving you a problem I'm not seeing?

hardbap:
Duh. Oops sorry about that. I'll close it out

chanks:
No problem!

PR #131, opened 2015-10-29
Locking algorithm somewhat more resistant to long-lived transactions

brandur:
Hey @chanks! I meant to try and send this in months ago, but never got around to it, so I wanted to loop back around and get your thoughts. This is my personal branch that's slightly more tolerant to long-lived transactions in a database as I think that you probably already remember from [an article and discussion on HN not too long ago](https://brandur.org/postgres-queues).

Possible considerations here:
- The branch works, but isn't perfect. Long-lived transactions will still take it down.
- This stuff is pretty hard to test, so I haven't included coverage of any substance here. (Let me know if you have any ideas on how to improve this.)
- There is some increase to code complexity and it may be surprising to people reading the codebase.

Let me know if you have any other thoughts/ideas here. Thanks!

/cc @mikehale

### DIFF
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -75,14 +75,14 @@ module Que
         new(:args => args).tap { |job| job.run(*args) }
       end
 
-      def work(queue = '')
+      def work(queue = '', min_job_id = 0)
         # Since we're taking session-level advisory locks, we have to hold the
         # same connection throughout the process of getting a job, working it,
         # deleting it, and removing the lock.
         return_value =
           Que.adapter.checkout do
             begin
-              if job = Que.execute(:lock_job, [queue]).first
+              if job = Que.execute(:lock_job, [queue, min_job_id]).first
                 # Edge case: It's possible for the lock_job query to have
                 # grabbed a job that's already been worked, if it took its MVCC
                 # snapshot while the job was processing, but didn't attempt the
--- a/lib/que/sql.rb
+++ b/lib/que/sql.rb
@@ -48,6 +48,7 @@ module Que
           FROM que_jobs AS j
           WHERE queue = $1::text
           AND run_at <= now()
+          AND job_id > $2::bigint
           ORDER BY priority, run_at, job_id
           LIMIT 1
         ) AS t1
--- a/lib/que/worker.rb
+++ b/lib/que/worker.rb
@@ -54,6 +54,16 @@ module Que
 
     private
 
+    # The interval in seconds between when the Que worker should perform a
+    # "hard lock" by throwing out the last job ID it worked and locking without
+    # using a predicate on `job_id`. The disadvantage of this is that the lock
+    # may take significantly longer if a long running transaction has caused a
+    # large number of dead tuples to accumulate in the table. The advantage is
+    # that it allows any jobs with IDs lower than any current worker is
+    # handling to be discovered and locked (that may happen in the case of a
+    # crashed worker or an out-of-order transaction commit).
+    HARD_LOCK_INTERVAL = 10
+
     # Sleep very briefly while waiting for a thread to get somewhere.
     def wait
       sleep 0.0001
@@ -70,8 +80,8 @@ module Que
     end
 
     def work_loop
+      last_hard_lock_at = Time.now
       min_job_id = 0
-      worked_count = 0
 
       loop do
         cycle = nil
@@ -100,18 +110,13 @@ module Que
             raise "Unknown Event: #{result[:event].inspect}"
           end
 
-          # Constraining workers to a `job_id` floor has a significant
-          # disadvantage that it's possible for a job to be starved out if it
-          # was inserted with a non-sequential job ID and all workers had
-          # already locked themselves to higher identifiers.
-          #
-          # This block attempts to work around that issue by simply not using
-          # the `job_id` floor every so often, which allows any jobs that may
-          # have been missed to be picked up by doing things the hard way.
-          worked_count += 1
-          if worked_count >= 100
+          # Every so often try to take a lock the "hard" way, which is to say,
+          # without a predicate on `job_id`. This may help to discover jobs
+          # that have been bypassed due to a crashed worker or out-of-order
+          # transaction commit.
+          if HARD_LOCK_INTERVAL == 0 || Time.now > last_hard_lock_at + HARD_LOCK_INTERVAL
+            last_hard_lock_at = Time.now
             min_job_id = 0
-            worked_count = 0
           end
 
           Que.log(result)
--- a/lib/que/worker.rb
+++ b/lib/que/worker.rb
@@ -70,12 +70,15 @@ module Que
     end
 
     def work_loop
+      min_job_id = 0
+      worked_count = 0
+
       loop do
         cycle = nil
 
         if Que.mode == :async
           time   = Time.now
-          result = Job.work(queue)
+          result = Job.work(queue, min_job_id)
 
           case result[:event]
           when :job_unavailable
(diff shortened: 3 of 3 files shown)


chanks:
Hey, yeah, I'll definitely look at this. Unfortunately, while my work does have open-source Fridays, this one was too busy for me to take a look at this :( but I definitely will soon.

Just out of curiosity, have you run this in production and/or benchmarked it? I don't remember all the details around this off the top of my head, but I am pretty wary of anything that complicates the job lock query more, for fear of slowing it down.

brandur:
> Hey, yeah, I'll definitely look at this. Unfortunately, while my work does have open-source Fridays, this one was too busy for me to take a look at this :( but I definitely will soon.

@chanks Understood :/ I wouldn't say there's any particular rush here (Heroku is already using a branch).

> Just out of curiosity, have you run this in production and/or benchmarked it? I don't remember all the details around this off the top of my head, but I am pretty wary of anything that complicates the job lock query more, for fear of slowing it down.

It's been running in production on Heroku's core orchestrator for roughly six months now, and can probably be considered reasonably safe.

There's some benchmarks attached to an article [I wrote on the subject here](https://brandur.org/postgres-queues). In particular, here's [without the patch](https://brandur.org/assets/postgres-queues/pre-lock-time.png) and here's [with the patch](https://brandur.org/assets/postgres-queues/post-lock-time.png). Sorry the graphs kind of suck, but you can see that the before and after the non-degraded lock time is roughly on the same scale of time. I have a small test suite attached there as well that should allow us to extract more detailed metrics (if we want to) as well.

joevandyk:
Would PostgreSQL 9.5's SKIP LOCKED feature help with this problem at all?

brandur:
@joevandyk Good question. I'm not going to rule it out, but in general the problem described here isn't caused by having to iterate rows locked by another process, but rather having to iterate rows that are "dead" and continue to accumulate outside the visibility of any ongoing workers.

joevandyk:
Yeah, that makes sense.

Thanks for looking into this by the way. We ran into the *exact* same
problem about eight months ago. The fix was moving analytics queries to
redshift and adding more timeouts so we never had any long-running
transactions.

Joe

brandur:
> Thanks for looking into this by the way. We ran into the *exact* same problem about eight months ago. The fix was moving analytics queries to redshift and adding more timeouts so we never had any long-running transactions.

Essentially our solution as well! Good to hear that we weren't the only ones to run into this though :)

joevandyk:
should this be merged?

chanks:
I don't know yet, I haven't taken the time to dive into how it works, and I'm wary of doing anything to modify the job lock query until I do. I do think that keeping transactions short is generally the right way to do things, but I suppose that's not always possible.

mikehale:
> I don't know yet, I haven't taken the time to dive into how it works, and I'm wary of doing anything to modify the job lock query until I do. 

This article provides a great in-depth analysis of the problem and the solution we used (this PR): https://brandur.org/postgres-queues.

> I do think that keeping transactions short is generally the right way to do things, but I suppose that's not always possible.

True. We have been actively working towards short transactions, but it is a process and having this solution in our toolbelt has been invaluable over the past 12+ months. An additional case where this turned out to be very helpful was when adding indexes to a large table in the same database. The schema changes took > 1h, and during that time it caused the `que_jobs` table to have > 50k dead tuples. However, throughout the process the median lock time for `que_jobs` stayed under 0.4 seconds.

chanks:
Been thinking this over, and graphing out example data to try to figure out how I think this change would affect operations on various data sets.

The danger in using job_id as a filter (only looking for jobs with ids greater than the id of the last job you worked) is that valid jobs that happen to have low ids will tend to be penalized. It seems to me that the two groups these jobs fall into will be low-priority jobs in a large backlog, or jobs that were scheduled far in advance a long time ago. I guess you could make the argument that if any jobs are going to be penalized, it should be those two groups.

I'm curious about the reasoning in choosing job_id as a filter? It seems to me that filtering for either (priority, run_at, job_id) or (run_at, job_id) larger than that of the last job you worked should do the trick, and make better use of the index, too? I'm unsure whether the second one would make good use of the index (it might), but it'd hopefully protect against a worker that operated on a low-priority record being unable to see a high-priority job.

I also wonder about the behavior of having each worker do a "hard lock" every ten seconds, and what makes number of seconds per worker a good metric. Also, there's not currently a mechanism for a worker to do a "hard lock" immediately after being unable to find any jobs in a query, which seems to me like a good time to do it.

brandur:
Hi @chanks,

Sorry about the terrible delay in responsiveness here, unfortunately I let my procrastination get the better of me.

So just first of all, I left Heroku about a year ago and am no longer working day-to-day with Que, so I don't really have all that much skin in the game anymore. That leaves me without too much incentive to champion this through, so if there's some reluctance, I'm happy to close it.

Overall I'd still rate the approach here as a decent balance between being advantageous for larger production systems while also being "not that bad" as far as hacks go. The _real_ solution is for Postgres' heap scans to not result in such degenerate behavior, but I'm not sure that we can expect this anytime soon. Also, if you do have some old queries running on your big systems this patch will stave off disaster for a little while, but if left unchecked, they'll still take you down in the end.

> I'm curious about the reasoning in choosing job_id as a filter? It seems to me that filtering for either (priority, run_at, job_id) or (run_at, job_id) larger than that of the last job you worked should do the trick, and make better use of the index, too? I'm unsure whether the second one would make good use of the index (it might), but it'd hopefully protect against a worker that operated on a low-priority record being unable to see a high-priority job.

Honestly, I can't recall exactly, but I think the reason was just because it doesn't matter all that much. If you're able to follow any index condition at all, then that's good enough for this patch to be somewhat helpful. I probably left out the other index fields just for the sake of simplicity.

> I also wonder about the behavior of having each worker do a "hard lock" every ten seconds, and what makes number of seconds per worker a good metric. Also, there's not currently a mechanism for a worker to do a "hard lock" immediately after being unable to find any jobs in a query, which seems to me like a good time to do it.

Once again, there isn't a huge amount of justification for this exact implementation compared to alternatives, but the basic reasoning is that these sorts of lost jobs will hopefully be quite rare (if you have a lot of workers crashing then you're likely to have other sorts of trouble); as long as there is some kind of mechanism to occasionally go back through and pick these up, then it's probably adequate for most uses. This of course wouldn't hold up if your workload requires some kind of mission critical SLA on how quickly every job must get worked, but if you had that sort of requirement you probably wouldn't be using this sort of system in the first place.

chanks:
Thanks a lot, and don't worry about the delay, we're all very busy 😄 and
it's nothing compared to my delay in getting to this patch in the first
place.

I think it's obviously something important to consider, so let's leave this
open for now. I'd like to try to recreate the problem in a benchmark and
try the various query tweaks to fix it before committing anything for it.
Unfortunately, I still don't know when I'll have the time to do that, but
luckily I'm now finally using Que at my day job, so it could certainly
happen!

Thanks again for your time and effort and the report.

joevandyk:
I've ran into the lock function taking a lot of time recently when I had a lot of jobs in the table and I updated the `queue` column of a large number of them at once. While postgresql was doing the update, the lock function took a long time.

brandur:
Cool! Thanks for the quick follow up @chanks. Let me know if there's any other information or help here that I can provide.

siegy22:
@brandur Hey! I didn't read the whole PR.. because it's just huge 😂 

Do you think this is still a thing which should be merged as of now? 
I'm just sceptical because this PR has been open for 4 years. 😅

brandur:
> Do you think this is still a thing which should be merged as of now?

@siegy22 I think other mitigations have since been introduced. Regardless, I won't pursue this one any further so I'm closing it out.

PR #136, opened 2015-11-27
Warn about concurrency and plain PG connection

tpitale:


chanks:
It hadn't even occurred to me when trying to diagnose your problem that you'd be using the single PG connection. I think I only added it in the first place so that the range of adapters would be more complete, and not because it'd actually really be useful to anyone.

Do you have a good reason for using a single connection rather than a pool? If not, I'm tempted to just remove that section from the docs entirely.

tpitale:
I switched to a connection. I think removing it as an example is a good idea. I would replace or promote the PG connection in a simple pool as the supported solution if you don't use AR.

I was using it because my project doesn't use AR.

tpitale:
*connection pool*

chanks:
Alright, removed. Sorry that was such a pain.

tpitale:
Oh, not your fault at all. Thanks for all the help!

PR #138, opened 2015-12-10
Allow environment variables for configuration in bin/que

chendo:
This PR allows `QUE_QUE`, `QUE_WORKER_COUNT`, `QUE_WAKE_INTERVAL` and `QUE_LOG_LEVEL` to be used. Precedence order is command line options, environment variables, `Que.[var]`, then the specified defaults.

We were lucky and had the same `DB_POOL` to the default worker count when we transitioned to `bin/que`.

### DIFF
--- a/bin/que
+++ b/bin/que
@@ -59,7 +59,7 @@ end
 Que.logger ||= Logger.new(STDOUT)
 
 begin
-  if log_level = options.log_level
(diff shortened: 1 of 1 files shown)


chendo:
Build failure looks like it's the same failure as master

chanks:
Looks good, thanks!

Issue #143, opened 2016-02-27
Rails 5 issues


ollym:
```
{"class":"NoMethodError","message":"undefined method `before_committed!' for #\u003cQue::Adapters::ActiveRecord::TransactionCallback:0x007fe849b993b0\u003e"}
```

This doesn't when `Que.mode = :sync`

chanks:
That's interesting. I didn't write the integration with ActiveRecord's transaction logic, and really don't use Rails at all anymore, but I'd certainly accept a patch if anyone wants to take a crack at this. Though I imagine it would better belong in the dedicated que-rails gem, if that ever gets off the ground.

joevandyk:
https://github.com/chanks/que/pull/144

ollym:
Thanks, @joevandyk 

Until this makes it as a release, here is a quick fix as an initilizer:

```ruby
# config/initilizers/fix_que_for_rails5.rb

class Que::Adapters::ActiveRecord::TransactionCallback

  def before_committed!(*)
    # no-op
  end
end
```

chanks:
#144 merged!

PR #144, opened 2016-02-27
make compatible with rails 5

joevandyk:


pcantrell:
Confirmed that this fixes #143 on Rails 5.0.0.beta3.

PR #145, opened 2016-03-23
[Proposal] Add retry limit global configuration

tpitale:
* Que.retry_limit, with default to 100
* Lock job query uses retry_limit against error_count
* Update docs


tpitale:
Looks like the build is failing on master already: https://travis-ci.org/chanks/que/jobs/113501122.

joevandyk:
https://github.com/bundler/bundler/issues/3558 looks related

tpitale:
Based on comment on #106 it appears that the suggested method for this case may simply be to add a custom `handle_error` which moves or destroys the job when it should no longer be retried.

I may open another PR with more docs added to this area:

https://github.com/chanks/que/blob/master/docs/customizing_que.md#not-retrying-certain-failed-jobs

PR #153, opened 2016-05-11
Docs index via README

MyklClason:
I'm looking at using Que in a web app I'm working on, but I see this project doesn't have a wiki and there's no index for the docs. This makes it a bit difficult for me to find what I'm looking for, so I started a basic index. It's a list of links to each doc with a sublist of with any subheadings for that doc.

I'd recommend adding short descriptions (a sentence or two perhaps) to the index to make it a bit clearer as to what is covered in each topic.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,33 @@
+Docs Index
+==============
+
+- [Advanced Setup](docs/advanced_setup.md)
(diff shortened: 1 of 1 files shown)


chanks:
Thanks! I do think it's helpful to have something show up when browsing to the docs directory on Github.

PR #155, opened 2016-05-14
Fix relative URLs in docs index #153

nnc:
This just removes the extraneous `docs/` prefix in all URLs, as previously all links resulted in 404 when browser to on GitHub.


chanks:
:+1:

PR #157, opened 2016-06-02
Update docs index to use anchor links

MyklClason:
I've updated the index to use anchor links so that the links go directly to the (sub)heading rather than just the top of the page.


chanks:
Thanks!

PR #160, opened 2016-06-28
Handle nested transactions in Rails 5

greysteil:
Currently, the setup added to the spec in the PR causes a `NoMethodError: 'undefined method add_to_transaction'` when running against ActiveRecord 5.

This PR adds a naive fix, which makes Que compatible with ActiveRecord 5. It's not the best possible solution (which is described in a comment), but it removes the blocker for anyone moving to Rails 5.

### DIFF
--- a/lib/que/adapters/active_record.rb
+++ b/lib/que/adapters/active_record.rb
@@ -46,6 +46,14 @@ module Que
         def before_committed!(*)
           # no-op
         end
+
+        def add_to_transaction
+          # no-op.
+          #
+          # This is called when we're in a nested transaction. Ideally we would
+          # `wake!` when the outer transaction gets committed, but that would be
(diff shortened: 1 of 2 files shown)


greysteil:
@chanks - would be awesome if you could take a quick look over this now Rails 5 is out. Am I right that calling `wake!` here would be a nice to have, and needn't be a blocker for people updating?

chanks:
I'll be able to look at it tomorrow, Friday is my open source day at work.

chanks:
Looked at it, yeah, you're right, calling wake! would be a nice-to-have, but I don't think it's strictly necessary. And I'd worry about maintaining backward compatibility with older versions of Rails with more invasive changes.

Will merge and release a new minor version.

PR #167, opened 2016-07-26
Make Que's ActiveRecord adapter compatible with Rails 5.0

bramski:
This is about the most surgical fix I could think of which would allow reentrant on the checkout function without a massive refactor of the way that Que's SQL execution works.


chanks:
Thanks for the PR, I want to do a bit more research on how Rails 5.0 changed the connection logic before merging. My open-source day at work is Friday, I think I can get to it then.

bramski:
This really isn't working well actually.  There seems to be a few underlying problems in master which are causing the cleanup not to happen.  I think that the worker needs to be refactored so that it's not re-entrant.

Issue #147, opened 2016-03-23
[Proposal] Per-class error_handler


tpitale:
@chanks @joevandyk How would you feel about me sending a PR for per-class `error_handler`?

Spiked an implementation like (inside job.rb):

```ruby
if job
  if klass && klass.respond_to?(:error_handler)
    klass.error_handler(error, job)
  else
    count    = job[:error_count].to_i + 1
    interval = klass && klass.respond_to?(:retry_interval) && klass.retry_interval || retry_interval
    delay    = interval.respond_to?(:call) ? interval.call(count) : interval
    message  = "#{error.message}\n#{error.backtrace.join("\n")}"
    Que.execute :set_error, [count, delay, message] + job.values_at(:queue, :priority, :run_at, :job_id)
  end
end
```

The only question I have is determining if the job should be destroyed or left up to the method to ensure that the job is destroyed. Seems risky to leave jobs hanging around.

I could add something so that `Que.execute :set_error` just sets the error and not a new `run_at`. And then change the current usage to be something like `Que.execute :set_error_and_retry`.

chanks:
I made a branch a while ago (https://github.com/chanks/que/compare/error-handling) to support a handle_error method in classes. This feels better to me than the class-level configuration used so far in Que because it's more composable - you can implement whatever error logic you want in a module and then include it in whatever job classes you want it to apply to (or just include it in Que::Job directly, for that matter). You can see the API I was thinking of in this comment: https://github.com/chanks/que/pull/106#issuecomment-94067506

I haven't put more work into it because nobody else seemed really invested in it and I don't have any projects using Que right now, so it hasn't been a priority for me (though I expect that to change in the next month or two :smile:). But does that sort of API (where an instance method would be responsible for deciding whether to destroy or retry the job) seem reasonable and useful to you?

tpitale:
That looks good! If you would rather go in that direction, I'm happy to close this.

chanks:
I think it makes more sense? But I haven't actually tried it, and I think
that branch is out of date. Friday is my open source day at work, I can try
to look into it then. I'm kind of interested in other critiques before
settling on it, though - there was conversation about it in another issue
that stalled out, as I recall. But at least I can bring it up to date and
give you something to try out, if this is a problem you're needing to solve
soon.

tpitale:
Yeah, I'm not in a huge rush. I'd say bring the branch up-to-date, and open for discussion.

chanks:
[...]

PR #171, opened 2016-09-09
Spike doc changes for error handling with per-job handle_error method

tpitale:
Related to #147.


chanks:
Closing old issues. We have documentation for this feature now - I guess we didn't when you submitted this? Anyway, looks like they both cover all the same points, but if there's anything to add in the future please let me know!

PR #182, opened 2017-01-05
Set default queue to 'default', fix ActiveJob and devise-async not wo…

ryancheung:
…rking bug

This fixes devise async mail not sending bug in :async mode.


ryancheung:
if job was inherited from ActiveJob::Base, job also would not be executed in :async mode

PR #190, opened 2017-06-01
Only call parse on JSON_MODULE if it's defined

rwz:
This addresses https://github.com/chanks/que/issues/189

### DIFF
--- a/lib/que/adapters/base.rb
+++ b/lib/que/adapters/base.rb
@@ -102,7 +102,11 @@ module Que
       CAST_PROCS[1184] = Time.method(:parse)
 
       # JSON.
-      CAST_PROCS[114] = -> (value) { JSON_MODULE.parse(value, create_additions: false) }
+      CAST_PROCS[114] = if JSON_MODULE.respond_to?(:parse)
+        -> (value) { JSON_MODULE.parse(value, create_additions: false) }
+      else
+        JSON_MODULE.method(:load)
+      end
 
       # Boolean:
       CAST_PROCS[16] = 't'.method(:==)


chanks:
Hey, thanks @rwz. @hmac, who pointed out the issue initially, also realized that there'd be a problem with MultiJson. I'm trying to figure out whether it's safe to just always call `JSON_MODULE.load(json, create_additions: false)`

chanks:
Yeah, it seems that way to me. I just pushed a fix to master, would you verify that it works for you? If it does I'll cut a new release.

rwz:
This is very likely to fix the issue for our codebase that has MultiJson in it. Seems like we'd also wanna someone with stdlib JSON to weight in. Or... I guess all the specs are exercising no-MultiJson case, so it's probably a good enough warranty?

chanks:
Yeah, that's my thinking, I'm not worried about the stdlib case. Is it convenient for you to test this, just to double-check that MultiJson works? If not I suppose I'll just release - it's a pretty simple fix, seems to me.

While we're here, I'm wondering if there's a compelling use-case anymore for Que to use MultiJson at all? Way back when it was introduced I believe the rationale was that stdlib had poor performance, but I'm not sure that that's true these days? I'm kind of inclined to remove support (in a major version bump) for simplicity's sake.

rwz:
Yeah, our CI's just verified that master works fine.

chanks:
Alright, released 0.12.3.

rwz:
Cool, thanks!

chanks:
Hey @rwz, I just pushed a change to master to drop support for MultiJson, since it doesn't seem to be necessary anymore and removing it simplifies things (see #191). When it's convenient would you check and make sure that the current master works for you?

Thanks!

rwz:
I can check now. Will report back in a few.

rwz:
We're green.

PR #192, opened 2017-06-12
explicitly require ActiveRecord::Base

mikz:
Otherwise it is possible to get exception when running tests: `test/test_helper.rb:23:in `<class:TestCase>': undefined method `fixtures' for ActiveSupport::TestCase:Class (NoMethodError)`

This adapter uses ActiveRecord::Base so it is only fair to require it.

### DIFF
--- a/lib/que/adapters/active_record.rb
+++ b/lib/que/adapters/active_record.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'active_record/base'
+
 module Que
   module Adapters
     class ActiveRecord < Base


chanks:
Hmm, generally the ActiveRecord adapter is only loaded when ActiveRecord is explicitly set as the connection source, which would mean that ActiveRecord::Base should already be loaded. I'm also not sure why you'd get that error - Que itself shouldn't have anything to do with fixtures.

Are you doing something unusual like requiring the adapter yourself?

mikz:
Well Que does start workers in the background that start lazy loading stuff. 
It is Rails API application, that does not have all the frameworks: https://github.com/3scale/zync/blob/ea851aff8a070fb9527afab114bd99d15ac4c849/config/application.rb#L8

Fixtures are loaded as part of ActiveRecord::TestCase and due to the autoloading (I guess) they are not loaded yet when test helper tries to configure them: https://github.com/3scale/zync/blob/ea851aff8a070fb9527afab114bd99d15ac4c849/test/test_helper.rb#L32

I have other complains about the Rails integration, but will open another issue about that (running workers inside rake tasks, not waiting for after_initialize, etc.).

chanks:
Closing old issues. I'm not sure what was going on here, but it doesn't seem like anyone ran into it, and the ActiveRecord integration has been completely rewritten since then. Please reopen if you still have problems on the 1.0 beta (master).

PR #200, opened 2017-12-03
Add reference to que-scheduler

hlascelles:
This adds a reference to [que-scheduler](https://github.com/hlascelles/que-scheduler), and a summary in the recurring jobs section.


chanks:
I kept the entry in the list of related gems, but felt like the description in the customization doc was kind of out of scope (it feels weird for the docs to be describing other gems), so I removed it. Thank you though!

hlascelles:
:+1:

PR #201, opened 2017-12-25
Fix ActiveRecord migration error

### LINKED ISSUE #194: Fresh install fails
I believe because of something introduced in Rails 5.1

```
StandardError: An error has occurred, this and all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class AddQue < ActiveRecord::Migration[4.2]
```

NARKOZ:
>Directly inheriting from ActiveRecord::Migration is not supported.
Please specify the Rails release the migration was written for:

>  class AddQue < ActiveRecord::Migration[4.2]

Fixes #194

### DIFF
--- a/lib/generators/que/templates/add_que.rb
+++ b/lib/generators/que/templates/add_que.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class AddQue < ActiveRecord::Migration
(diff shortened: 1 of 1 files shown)


chanks:
Merged, thank you!

PR #204, opened 2018-02-20
Adds instructions for Sequel migrator

davidpelaez:
This minor documentation change explains how to create a Sequel migration for Que. It includes both the up and down methods so it can be safely rolled back. The current documentation includes a generator for ActiveRecord but there's no information of the equivalent when using Sequel.

### DIFF
--- a/docs/using_sequel.md
+++ b/docs/using_sequel.md
@@ -7,6 +7,22 @@ DB = Sequel.connect(ENV['DATABASE_URL'])
 Que.connection = DB
 ```
 
+If you are using Sequels' migrator, you can use the following migration to create the required tables:
+
+```ruby
+require 'que'
+Sequel.migration do
+  up do
+    Que.connection = self
+    Que.migrate! :version => 3
+  end
+  down do
+    Que.connection = self
+    Que.migrate! :version => 0
+  end
+end
+```
+
 Then you can safely use the same database object to transactionally protect your jobs:
 
 ```ruby


chanks:
Not against including this in the docs, but I'm a bit skeptical of the `require 'que'` and the `Que.connection = self` calls. Every time I've used Que with Sequel I've always required it and established the connection as part of the app boot (after requiring 'sequel' and establishing the DB connection in the first place), which seems pretty reasonable to me. Is there a reason you're setting it up late like this?

davidpelaez:
@chanks this is because I do not boot the app when using the migrations because it's possible that your app code and migrations have different expectations, the code is usually ahead of the migrations and the migrations need to be run for the code to work. 

If you are using the Sequel migrator I think this is the simplest option. If you have a custom rake task then you can make the connection a constant that you can use in the migration:

```ruby
namespace :db do
  desc "Run migrations"
  task :migrate, [:version] do |t, args|
    require 'bundler'
    Bundler.setup :default
    require "sequel"
    Sequel.extension :migration
    # ***** the DB connection is now a constant ****
    DB = Sequel.connect(
      # Default value for testing
      ENV.fetch("DATABASE_URL", "postgresql://localhost:5432/app_test")
    )
    if args[:version]
      puts "Migrating to version #{args[:version]}"
      Sequel::Migrator.run(DB, "db/migrations", target: args[:version].to_i)
    else
      puts "Migrating to latest"
      Sequel::Migrator.run(DB, "db/migrations")
    end
  end
end
```

You can either setup the connection there (a little weird to me because it separates the setup from the migration where it's used which would rarely be more than one) or write something like `Que.connection = DB` but you need to make sure of course that DB was passed to the migrator correctly. Until now the connection was a local variable making sure I was not interacting with the DB more than needed, instead I would only use `from(:table_name)` to interact with tables but never had the need for the full connection so that's why I used self.

Should we change it then and also display the rake task info? As I said before this would not work if you are using the built in command line migrator bundled with Sequel. Something like this would not work: `sequel -m db/migrations postgres://localhost/sequel_test` because you wouldn't have access to the setup to name `DB`, that's why the proposed way seems more universal to me.

chanks:
Sorry for the delay, I'm just cleaning up some old issues now. Thanks!

PR #205, opened 2018-02-22
Rails railtie no longer barfs due to "uninitialized constant"

harukizaemon:


joevandyk:
rails autoload is so annoying

chanks:
:+1:

PR #208, opened 2018-03-02
Handle errors with blank message when setting last_error.

joehorsnell:
If an error is raised with a blank message, as (for example) sometimes
`OpenSSL::Cipher::CipherError` can be, the `last_error` that is saved
to the DB for that job has a confusing first blank line. For older
versions of Que (before 3feebc56738a3b114472386a4c68e3ac1b541a4c) with
no stacktrace, all you get is an empty string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Unreleased
+
+*   Fix handling of last_error message for errors with blank message.
+
 ### 0.14.2 (2018-01-05)
 
 *   Deprecate the Que.disable_prepared_statements= accessors.
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -33,6 +33,10 @@ module Que
       @attrs[:error_count]
     end
 
+    def error_message
+      self.class.send(:error_message, @_error)
+    end
+
     def handle_error(error)
       error_count = @attrs[:error_count] += 1
       retry_interval = self.class.retry_interval || Job.retry_interval
@@ -41,7 +45,6 @@ module Que
     end
 
     def retry_in(period)
-      error_message = "#{@_error.message}\n#{@_error.backtrace.join("\n")}"
       Que.execute :set_error, [period, error_message] + @attrs.values_at(:queue, :priority, :run_at, :job_id)
       @retried = true
     end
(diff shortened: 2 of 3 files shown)


joehorsnell:
I don't _think_ those Travis failures are related to my changes, @chanks?

chanks:
Thanks @joehorsnell! I tweaked the logic a bit to just always include the class name of the error, so that should be more useful in general. I also threw in a change to truncate the error message to 500 characters. These are both things I did in the 1.0 branch a while ago.

And yeah, those spec failures aren't your fault. The master branch (0.x) has a lot of iffy tests, 1.0 is a lot better about that.

joehorsnell:
You are welcome, @chanks! That is service for you - I only opened the PR a few hours ago and you've already published a new [gem version](https://rubygems.org/gems/que/versions/0.14.3)! Thanks vm. FYI, couple of minor comments on your [master commit](06d7712fb4d3555d19089096b19ed6e669338bfd).

Ah yes - as you say, I can see [here](2403036f0242d8d809c317862dec6e0a0a550fd6) that you did pretty much the exact same fix as this PR on the 1.0 branch. Just OOI, what's your plan with 1.0 in terms of compatibility (or lack of) with 0.x, upgrade path from 0.x and any rough timeline for when you are aiming to release it?

Thanks for your efforts on Que by the way. We've been gradually migrating our services at [Bamboo Loans](https://www.bambooloans.com/) from using Resque to Que - such an improvement - not losing jobs when workers die, for example!

joehorsnell:
You're quick - you'd already replied to my other comments! 😆

joehorsnell:
Actually, I can see you've got a quite extensive [CHANGELOG](/chanks/que/blob/fb94321b40f763fb48db1a63f0c1084bee961388/CHANGELOG.1.0.beta.md) with the 1.0 updates in it, so I'll take a look. Seems to be a pretty major overhaul/re-write - nice work.

chanks:
Yup! Upgrading from 0.14.3 to the beta is pretty straightforward - there's a new migration to run and some API changes, but that's pretty much it.

I don't know about when it'll be released - there are a couple outstanding things with the beta that I'm tracking down as I have time (mostly on Fridays, that's open-source day at my work). I'd like for the next release to be a release candidate. But I think the beta is pretty reliable - we have a new product at my day job that's been running it in (early) production for a few months.

PR #215, opened 2018-04-19
disable run_synchronously in Rails tests

### LINKED ISSUE #214: undefined method `run_asynchronously=' for Que:Module
[This line](https://github.com/chanks/que/blob/fb94321b40f763fb48db1a63f0c1084bee961388/lib/que/rails/railtie.rb#L8) calls a function `Que.run_asynchronously=` that doesn't exist.

bgentry:
This is my attempt to fix #214. Hoping that I correctly understood the intent of this bit of code.

Fixes #214.

### DIFF
--- a/lib/que/rails/railtie.rb
+++ b/lib/que/rails/railtie.rb
@@ -5,7 +5,7 @@ module Que
     class Railtie < ::Rails::Railtie
       config.que = Que
 
-      Que.run_asynchronously = true if ::Rails.env.test?
+      Que.run_synchronously = false if ::Rails.env.test?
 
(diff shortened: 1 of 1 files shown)


bgentry:
Closing as you indicated in #214 that the correct fix was to just delete this line.

chanks:
The original intent was to enable synchronous job running in the test environment - I just flubbed and typed run_asynchronously instead of run_synchronously. Like I mentioned in #214 I don't feel like that's a best practice anymore, but I'm certainly open to other perspectives on it, if you think it'd be useful?

Either way, thanks for bringing it to my attention!

PR #226, opened 2018-08-14
Remove public schema references and defer to search_path

nbarthel:
I use multiple schemas and the hard-coded references to public break the migrations.

It's late and I'm tired and must admit I haven't run the specs, but assume that schema_path settings will kick in...

### DIFF
--- a/lib/que/migrations/4/up.sql
+++ b/lib/que/migrations/4/up.sql
@@ -81,7 +81,7 @@ ALTER TABLE que_jobs
         AND
         (jsonb_array_length(data->'tags') <= 5)
         AND
-        (public.que_validate_tags(data->'tags'))
+        (que_validate_tags(data->'tags'))
       )
     )
   ),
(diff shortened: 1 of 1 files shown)


chanks:
Putting the job table in the public schema specifically was intentional, because having more than one job table in the same DB would be problematic (you can't scope advisory locks, so they'll interfere with each other, and polling becomes impractical with N job tables anyway). Not to mention that this migration has already been included in a gem release, so editing it after the fact would probably cause some problems.

If you're using the approach of running the same set of migrations against multiple schemas, I don't think there's anything wrong with wrapping the `Que.migrate` call in a conditional, so that it only runs once - I've worked on projects that did that, and it didn't cause any long-term problems.

nbarthel:
Thanks for the comments!  We have no public schema, not that we couldn't, but will consider going this route.  I dug a little deeper and realized the extent to which you had committed to public.

Issue #127, opened 2015-10-22
Document ActiveJob compatibility/configuration


bensheldon:
I eagerly recommend Que to everyone, but the first question they ask is "Is it ActiveJob compatible?". The answer is Yes, but ActiveJob is not mentioned in the Readme nor is there an example in the docs directory. I'd be happy to submit a PR to improve the documentation for using Que with ActiveJob if this is an unintentional omission.

chanks:
In the 1.0 branch I've actually removed everything that's Rails-specific (the Railtie, the migration generator, etc). Those parts have always been the most poorly tested and the most time consuming to maintain, especially since I don't personally use Que with Rails (My day job uses Rails + Sidekiq, and I have personal projects that use Que with Roda and Sinatra). I've also gotten frustrated with ActiveJob in the past, and I've grown kind of skeptical that a job queue is the right place for this kind of plug-and-play abstraction.

I've actually been considering whether it would be good to have a disclaimer on the Readme, saying that if the reader wants a quick and easy integration with Rails that they don't have to think too much about, they should probably be looking elsewhere. I'm happy if other people get use out of Que, and when I first launched it I was certainly hoping for a significant level of adoption, which is why I included the Rails integration in the first place. But these days I don't really feel like a large userbase for it is a goal I need to strive toward.

I think it'd be nice if somebody with more of an interest in the Rails stack maintained a gem that promised an easy integration with Que, and I'd certainly try to keep Que flexible enough to make their job easy. But I'm not really plugged into the Rails Way anymore, and I can't really justify the time expenditure to support everyone that is.

chanks:
BTW, sorry if that reads as brusque. I don't mean to brush aside your concerns, I'm just laying out how I feel. Does my wanting to drop closer Rails support seem unreasonable to you, or anyone else reading?

bensheldon:
@chanks that reads completely polite :smile: 

I might disagree that "a large userbase" and a small, focused, opinionated tool like Que are mutually exclusive. But I can agree and sympathize about externalizing the Rails integration into a separate gem if that's not where you want to spend time in development or support. As a middleground, you might consider still maintaining control (e.g. owning the repo and Rubygem) of a potential `que-rails` gem but offload the support to co-maintainers. Just my preference, but I'd rather have an officially blessed gem than multiple projects in various states of abandonment on Github. 

Do you have a public timeline/milestone for the 1.0 release? I hope there can be more community discussion and input; not to change your mind, but at least to pick up the work to be done.

[...]

PR #231, opened 2018-10-15
Mention Active Job integration in the README

giovannibonetti:
Thank you for this great library!

I've read the discussion on #127 and it seems like now there is great Active Job integration, so maybe it would be nice to document it so that new users can easily find it and decide to use it.

It took me a while to find this library because it is not mentioned in the [Rails Guides](https://guides.rubyonrails.org/active_job_basics.html#starting-the-backend), even though it is mentioned in the [API docs](https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters.html). It seems like the guides are more beginner-friendly and have links specific to the README section explaining how to integrate the library with Active Job. Hence if we have that section in the README we could have a link from Rails Guides pointing directly to it which would bring people here.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,24 +126,22 @@ usage: que [options] [file/to/require] ...
 
 You may need to pass que a file path to require so that it can load your app. Que will automatically load `config/environment.rb` if it exists, so you shouldn't need an argument if you're using Rails.
 
-## Usage with Active Job
+## Additional Rails-specific Setup
 
-If you would rather integrate Que with Active Job, you can do it by setting the adapter in `config/application.rb` or in a specific environment by setting it in `config/environments/production.rb`, for example:
+If you're using ActiveRecord to dump your database's schema, please [set your schema_format to :sql](http://guides.rubyonrails.org/migrations.html#types-of-schema-dumps) so that Que's table structure is managed correctly. This is a good idea regardless, as the `:ruby` schema format doesn't support many of PostgreSQL's advanced features.
+
+Pre-1.0, the default queue name needed to be configured in order for Que to work out of the box with Rails. In 1.0 the default queue name is now 'default', as Rails expects, but when Rails enqueues some types of jobs it may try to use another queue name that isn't worked by default - in particular, ActionMailer uses a queue named 'mailers' by default, so in your app config you'll also need to set `config.action_mailer.deliver_later_queue_name = 'default'` if you're using ActionMailer.
+
+Also, if you would like to integrate Que with Active Job, you can do it by setting the adapter in `config/application.rb` or in a specific environment by setting it in `config/environments/production.rb`, for example:
 ```ruby
 config.active_job.queue_adapter = :que
 ```
 
 Que will automatically pick up the database from Active Record, so there is no need to configure anything else.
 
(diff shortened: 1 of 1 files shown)


siegy22:
@chanks ping

siegy22:
Hey there!

Que moved: https://github.com/que-rb/que
Can you reopen your PR on the repository there so we can have a look at it?

giovannibonetti:
Hi! It seems we are already on the new repository:
https://github.com/que-rb/que/pull/231

Doesn't it work the way it is now?

siegy22:
Yeah forget my comment, we first forked the repository but then decided to transfer it to the organization.

README.md
```diff
@@ -126,6 +126,19 @@ usage: que [options] [file/to/require] ...
 
 You may need to pass que a file path to require so that it can load your app. Que will automatically load `config/environment.rb` if it exists, so you shouldn't need an argument if you're using Rails.
 
+## Usage with Active Job
+
+If you would rather integrate Que with Active Job, you can do it by setting the adapter in `config/application.rb` or in a specific environment by setting it in `config/environments/production.rb`, for example:
+```ruby
+ActiveJob::Base.queue_adapter = :que
```
siegy22:
```suggestion
config.active_job.queue_adapter = :que
```

README.md
```diff
@@ -126,6 +126,19 @@ usage: que [options] [file/to/require] ...
...
+```
+
+Que will automatically pick up the database from Active Record, so there is no need to configure anything else.
```
siegy22: I think you mean `Active Job`, right?

README.md
```diff
@@ -126,6 +126,19 @@ usage: que [options] [file/to/require] ...
 
 You may need to pass que a file path to require so that it can load your app. Que will automatically load `config/environment.rb` if it exists, so you shouldn't need an argument if you're using Rails.
 
+## Usage with Active Job
```
siegy22: Can't we just use the "Additional Rails-specific Setup" for that?

README.md
giovannibonetti: When I read [this part of the code](https://github.com/que-rb/que/blob/master/lib/que.rb#L79-L84) I thought the database connection was managed by Active Record. In fact I think you can use Que with Active Record without using Active Job at all, but I might be wrong.

Does that make sense?

README.md
giovannibonetti: Done! Is that what you had in mind?

README.md
siegy22: Ah. Now I get it. This says that `que` will automatically use the Active Record connection.

I think the `pick up` is confusing. Can we use something like:

> Que will automatically use the database configuration of your rails application.

README.md
siegy22: Yes! 😄

README.md
siegy22: @giovannibonetti What do you think?

siegy22:
@giovannibonetti Maybe you missed my feedback on the second part. So see this message as a friendly "ping", because I'm really looking forward to have this in the README!

giovannibonetti:
All right, I didn't notice that part. I will change it now. 👍

siegy22:
I'm not a native english speaker, so if you disagree, feel free to keep it as is.

giovannibonetti:
I'm not either, but your suggestion looks good, so I think it is worth the change! 👍

README.md
giovannibonetti: Done. Anything else?

siegy22:
Thanks for your contribution! 💚

giovannibonetti:
You're welcome! 😀

PR #239, opened 2019-04-09
Fix typecast when value is already type of Time.

f-mer:
I'm running que `1.0.0.beta3` with rails `6.0.0.beta3` and ran into issues with typecasting. Seems like the value of `run_at` is already of type `Time` so trying to parse it again raises an error. 
This PR tries to fix the described issue 🙂

### DIFF
--- a/lib/que/connection.rb
+++ b/lib/que/connection.rb
@@ -152,7 +152,12 @@ module Que
       },
 
       # Timestamp with time zone
-      1184 => Time.method(:parse),
+      1184 => -> (value) {
+        case value
+        when String then Time.parse(value)
+        else value
+        end
+      }
     }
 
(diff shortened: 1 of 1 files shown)


chanks:
That's unusual that that's necessary for you... do you have any idea what's going on? Que borrows the underlying PG connections that ActiveRecord is using, so I wonder if Rails 6 has pushed its result typecasting to a lower level?

f-mer:
I tried to figure out what's going on but didn't find a peace of code that could have typecasted the value.
For now I just switched to resque because of some other appealing features. 

Feel free to close this pull request as I am no longer using this project.
Thanks a lot for your support and quick response! 

Have a nice day :slightly_smiling_face:

Issue #243, opened 2019-05-10
Recommend a periodic manual vacuum in the readme?


mikehale:
We have been running que-0.14.3 for awhile on a very high churn database. Recently we nearly had an outage because other large tables in our database were being autovacuumed which caused all 3 of the autovacuum workers to be unavailable for over 90 minutes, which meant that the que_jobs table was not getting vacuumed, which lead to working jobs more slowly than they were being added. Since we had plenty of IOPS available, to fix this issue and keep it from possibly happening again (if we got unlucky with autovacuum timing) we decided to add a process that started a manual vacuum of our high churn tables (of which que_jobs is one) every 5 minutes. So far this has been working great for us, and keeps our lock times at a very predictable level. I guess my question is, do you think periodic manual vacuums are something that should be generally recommended, and maybe included in the readme, or one of the other docs?

chanks:
Maybe? I don't think it's general enough to belong in the readme, and I also think that whenever your database gets busy enough that the vacuum settings become significant, you're probably at the point where you want a Postgres expert to be managing it, and they'd already know how to handle this sort of thing. But it couldn't hurt to throw it in the docs somewhere.

joevandyk:
I’ve also had to manually vacuum the jobs table regularly.

mikehale:
I'm not a postgres expert by any stretch, just a developer and operator. It would be nice to have a dedicated postgres expert, but we still aren't in a position to have a person for that, and I imagine a lot of others are in the same boat. We've been running que in production since 2015, and only now realized that not relying on autovacuum to hopefully be available when we need it to keep dead tuples from growing without bound is a good idea. I thought it might be a good thing to call out so others can learn from our experience.

I haven't looked into que 1x too deeply, do you think the same advice will apply there?

chanks:
I'm sure it will. I'd happily merge a PR adding this to the docs somewhere.

chanks:
Closed by #244

PR #244, opened 2019-05-10
Add doc on vacuuming

mikehale:
I wasn't sure how far to go with the example. The one I included depends on a now non-existant CronJob class that used to be in `docs/customizing_que.md`, but it should still give folks an idea of how to proceed. If you want I can simplify the example down to just `VACUUM VERBOSE ANALYZE que_jobs`, and say that it should be run about every 5 minutes by some process.

### DIFF
--- a/docs/vacuuming.md
+++ b/docs/vacuuming.md
@@ -1,0 +1,20 @@
+## Vacuuming
+
+Because the que_jobs table is "high churn" (lots of rows are being created and deleted), it needs to be vacuumed fairly frequently to keep the dead tuple count down otherwise [acquring a job to work will start taking longer and longer](https://brandur.org/postgres-queues).
+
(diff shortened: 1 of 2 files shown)


chanks:
LGTM, thanks!

Closes #243

docs/vacuuming.md
```diff
@@ -0,0 +1,20 @@
...
+
+In many cases postgres will vacuum these dead tuples automatically using autovacuum, so no intervention is required. However, if your database has a lot of other large tables that take hours for autovacuum to run on, it is possible that there won't be any autovacuum processes available within a reasonable amount of time. If that happens the dead tuple count on the que_jobs table will reach a point where it starts taking so long to acquire a job to work that the jobs are being added faster than they can be worked.
+
+In order to avoid this situation you can kick off a manual vacuum against the que_jobs table on a regular basis. This manual vacuum will be more aggressive than an autovacuum since by default it does not back-off and sleep, so you will want to make sure your server has enough disk I/O available to handle the vacuum + any autovacuums + your workload + some overhead. However, by keeping the interval between vacuums small you will also be limiting the amount of work to be done which will aleviate some of the afforementiond risk of I/O usage.
+
+Here is an example recurring manual vacuum job that assumes you are using Sequel:
+
+```
+class ManualVacuumJob < CronJob
```
bogdan, line 12: Where does `CronJob` come from?

mikehale: Good question. I don't remember the original details, but it looks like that should probably change to `Que::Job`

PR #248, opened 2019-05-30
Add 1.x-compatible related projects to README

### LINKED ISSUE #221: related projects
Thanks for Que, I recently started using it and like it so far!

The readme says https://github.com/chanks/que/blob/master/README.md#community-and-contributing

> like some of the related projects above

looks like the related projects list was removed in 1896af990d360ed4882aacefc5f1649750592ac6

Maybe there could be a related projects doc in /docs?

coffenbacher:
This section was taken out as part of 1896af9
Presumably due to the 0.x->1.x upgrade breaking compatibility.

This commit adds projects
That have been confirmed as working with 1.x.

Closes #221


coffenbacher:
Hi @siegy22 , what do you think?

PR #249, opened 2019-06-26
fix parser of time in convert_result method

Pepan:
There is trouble on macOS, Rails 6 RC1 with (PostgreSQL) 11.3
Timestamp "run_at" is already returned as Time, not as String which is currently expected.

- changed direct method call to lambda with guard and original method call
- handled situation when value is already of a Time class by adding guard


chanks:
Merged, thanks!

PR #250, opened 2019-07-05
Simplify Sequel adapter

janko:
I noticed the Sequel adapter can be simplified a bit, so I went ahead and did it. There are several types of changes here:

* use `[]` instead of `Sequel.qualify`
* use `!~` instead of `Sequel.~(a => b)`
* use combination of `|`, `&` and `=~` operators instead of `Sequel.|`
* use `pg_json_ops` Sequel extension for building JSONB expressions
* pass an operator symbol to `#inject`

I feel like this made the code more consistent and easier to read. Note that all these changes are purely cosmetic, the behaviour should stay unchanged.

### DIFF
--- a/lib/que/sequel/model.rb
+++ b/lib/que/sequel/model.rb
@@ -18,8 +18,8 @@ module Que
           subset :"not_#{name}", ~condition
         end
 
-        subset :ready,     conditions.values.map(&:~).inject{|a, b| a & b}
-        subset :not_ready, conditions.values.         inject{|a, b| a | b}
+        subset :ready,     conditions.values.map(&:~).inject(:&)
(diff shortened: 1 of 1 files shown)


siegy22:
Hey there!

Que moved: https://github.com/que-rb/que
Can you reopen your PR on the repository there so we can have a look at it?

janko:
Reopened – https://github.com/que-rb/que/pull/1

aried3r:
Just for the record, it "moved" again, but just to a an organization, I believe this can be reopened?

siegy22:
Yeah sorry, that's because we first forked it, and then transefered it. I'm gonna open a PR from his fork.

Issue #2, opened 2013-11-19
Coveralls, TravisCI & CodeClimate


BukhariH:
Might be wise to implement these three things!

They definitely do help keep development as clean and mishap free as possible! 

I believe the link for adding a codeclimate open source project is:
https://codeclimate.com/github/signup

PR #255, opened 2019-09-23
Setup GitHub actions

siegy22:
- Test against latest rails major versions (4.2, 5.2, 6.0)
  (the previous gemfile tested against rails 4.1, 4.2, 5.0, 5.1, 5.2.betaX)
- Test against postgres 9, 10, 11
- Test against ruby 2.5 and ruby 2.6
- Loosen bundler pinning to be able to use bundler 2.
- Add GitHub actions `ruby.yml` to do the testing.


siegy22:
Misclicked the target branch.. -> 
https://github.com/que-rb/que/pull/2

PR #256, opened 2019-09-23
Setup GitHub actions

siegy22:
- Test against latest rails major versions (4.2, 5.2, 6.0)
  (the previous gemfile tested against rails 4.1, 4.2, 5.0, 5.1, 5.2.betaX)
- Test against postgres 9, 10, 11
- Test against ruby 2.5 and ruby 2.6
- Loosen bundler pinning to be able to use bundler 2.
- Add GitHub actions `ruby.yml` to do the testing.
- Remove circleci configuration.


spec/spec_helper.rb
```diff
@@ -6,19 +6,15 @@
 # ActiveRecord and ActiveJob require ActiveSupport, which affects a bunch of
 # core classes and may change some behavior that we rely on, so only bring it in
 # in some spec runs.
-if ENV['USE_RAILS'] == 'true'
```
chanks: Just pointing out, the reason for `USE_RAILS` was to make sure that the gem still works if the application using it doesn't use ActiveSupport. In other words, to make sure that the tests will fail if any ActiveSupport-isms accidentally make it into the core gem.

The comment above explains this, but now it doesn't make any sense.

siegy22: Oh I totally missed that, I saw it in the circle ci configuration.

Should we add another dimension in the build matrix to test with and without this? On the other hand, this would make 36 build steps, what do you think?

siegy22: Instead of adding another dimension, we just test it once with `USE_RAILS` and once without. Probably easier to understand and reduces build time in general.

chanks: Yeah, that's probably fine! I went on a run of getting the test time down as much as I could a while ago - that's why the CircleCI logic did a lot of test runs in parallel on only a few machines. But whatever makes sense for Github actions.

siegy22: I think around 2 minutes are pretty reasonable.

Also the main argument was to still have some kind of idea about all the builds, if you have a matrix of 36 builds, have fun 😂

siegy22:
I'm going to merge it at this point, so we have testing on the pull requests.
Thanks for the feedback @chanks

PR #274, opened 2020-02-11
Set schema in table_name for ActiveRecord model

nikitug:
As long as `que_jobs` is put into `public` namespace by purpose, it would be good to explicitly set it in ActiveRecord model. Otherwise application with changed `search_path` is not able to access `que_jobs`.


siegy22:
So this only applies for postgres dbs which are setup with `public` removed in their search_path?

nikitug:
@siegy22 that's right

PR #275, opened 2020-02-27
Fix CI failure from Docker Postgres image

jonathanhefner:
The failure is caused by a change to the Docker Postgres base image.  Using an empty string for the database password (the default value) now causes initialization to fail, unless a specific environment variable is set.

See docker-library/postgres#681 for more information.


jonathanhefner:
See that the build failed with no changes: https://github.com/que-rb/que/actions/runs/45925992

siegy22:
Thank you for your contribution, I noticed the same behaviour in another project 😄

PR #276, opened 2020-02-27
Test with Ruby 2.7

jonathanhefner:
Rails 4.2 is excluded when testing with Ruby 2.7 because Ruby 2.7 breaks some Rails 4.2 internal code.  Note that Rails 4.2 is end-of-life, so these breakages won't be fixed.


siegy22:
Thanks for your contribution 💚

Issue #241, opened 2019-04-16
Worker stops executing jobs (deadlocks)


pawelniewie:
So far I don't have much details. We switched to que last week and noticed that every day some worker will stop executing new jobs. Restarting the process fixes the issue. I'm now in the process of figuring this out.

I'm using Que 1.x with Rails/ActiveJob, we have 3 workers running on 3 separate heroku dynos, each worker is running 20 threads, each worker has a pool of 20 connections + 1 for locker.

So far I noticed the worker will stop processing jobs after a few hours after booting up. Connections will be closed but threads will be still running.

pawelniewie:
```
# Turn on internal log
rbtrace -p 4 -e "Que.internal_logger = Logger.new('/tmp/que_internal.log')"

# See if there's anything in the logs
tail -f /tmp/que_internal.log

# Turn off internal log
rbtrace -p 4 -e "Que.internal_logger = nil"

# Clean up
rm /tmp/que_internal.log
```

Broken process logs nothing, others log correct events into the log.

pawelniewie:
Thread dump for a broken worker:

```
~ $ rbtrace -p 4 --backtraces
*** run `sudo sysctl kernel.msgmnb=1048576` to prevent losing events (currently: 16384 bytes)
*** attached to process 4
"eval:1:in `backtrace'
eval:1:in `block (2 levels) in parse'
eval:1:in `map'
eval:1:in `block in parse'
/app/vendor/bundle/ruby/2.6.0/bundler/gems/que-c7ae049db0ca/bin/command_line_interface.rb:238:in `sleep'
/app/vendor/bundle/ruby/2.6.0/bundler/gems/que-c7ae049db0ca/bin/command_line_interface.rb:238:in `block in parse'
/app/vendor/bundle/ruby/2.6.0/bundler/gems/que-c7ae049db0ca/bin/command_line_interface.rb:237:in `loop'
/app/vendor/bundle/ruby/2.6.0/bundler/gems/que-c7ae049db0ca/bin/command_line_interface.rb:237:in `parse'
/app/vendor/bundle/ruby/2.6.0/bundler/gems/que-c7ae049db0ca/bin/que:9:in `<top (required)>'
/app/vendor/bundle/ruby/2.6.0/bin/que:23:in `load'
/app/vendor/bundle/ruby/2.6.0/bin/que:23:in `<top (required)>'
[...]

PR #285, opened 2020-04-27
Do not mutate the JobBuffer from PriorityQueue to prevent deadlocks

gerrypower:
This PR addresses the deadlock issue #241 by not mutating the JobBuffer from within PriorityQueue.

### DIFF
--- a/lib/que/job_buffer.rb
+++ b/lib/que/job_buffer.rb
@@ -75,7 +75,7 @@ module Que
 
     def shift(priority = nil)
       queue = priority_queues.fetch(priority) { raise Error, "not a permitted priority! #{priority}" }
-      queue.pop
+      queue.pop || shift_job(priority)
     end
 
     def shift_job(priority = nil)
@@ -158,6 +158,10 @@ module Que
       sync { _stopping? }
     end
 
+    def job_available?(priority)
+      (job = @array.first) && job.priority_sufficient?(priority)
+    end
+
     private
 
     def _buffer_space
@@ -210,14 +214,13 @@ module Que
       def pop
         sync do
           loop do
-            return false if @stopping
-
-            if item = @items.pop
+            if @stopping
+              return false
+            elsif item = @items.pop
               return item
+            elsif job_buffer.job_available?(priority)
+              return false
             end
-
-            job = job_buffer.shift_job(priority)
-            return job unless job.nil? # False means we're stopping.
 
             @waiting += 1
             @cv.wait(mutex)


siegy22:
Thank you for your contribution! 

Is there any way to test this? Because we sure don't want to have a regression on that.

gerrypower:
Some test is required; do you have guidance on what you would like to see? It's a hard integration test, as I found it only deadlocked under load.

ebeigarts:
Here is an example test case:

```diff
--- a/lib/que/job_buffer.spec.rb
+++ b/lib/que/job_buffer.spec.rb
@@ -201,6 +201,50 @@ describe Que::JobBuffer do
     end
   end

+  describe "push and shift" do
+    it "should not deadlock" do
+      job_buffer # Pre-initialize to avoid race conditions.
+
+      concurrency = 4
+      jobs_count_per_thread = 10_000
+
+      push_thread = Thread.new do
+        (concurrency * jobs_count_per_thread).times do
+          job_buffer.push(*job_array[0])
+        end
+      end
+
+      shift_threads = concurrency.times.map do
+        Thread.new do
+          jobs_count_per_thread.times do
+            job_buffer.shift
+          end
+        end
+      end
+
+      deadlock_detected = false
+      unlock_thread = Thread.new do
+        sleep 10
+
+        deadlock_detected = true
+        push_thread.kill
+        shift_threads.each { |t| t.kill }
+      end
+
+      push_thread.join
+      (concurrency * jobs_count_per_thread).times do
+        job_buffer.push(*job_array[0])
+      end
+
+      shift_threads.each { |t| t.join }
+      unlock_thread.kill
+
+      if deadlock_detected
+        raise "Deadlock"
+      end
+    end
+  end
+
```

With this test case `master` fails almost on every run:

```
  1) Error:
Que::JobBuffer::push and shift#test_0001_should not deadlock:
RuntimeError: Deadlock
    que/lib/que/job_buffer.spec.rb:243:in `block (3 levels) in <top (required)>'
```

This PR seems to fix the issue, but on every second run it fails, I guess there is another race condition between `waiting_count` and `push` and inside `pq.push` there is `Que.assert(waiting_count > 0)`.

```
priority_queues.reverse_each do |_, pq|
  pq.waiting_count.times do
    job = _shift_job(pq.priority)
    break if job.nil? # False would mean we're stopping.
    pq.push(job)
  end
end
```

```
  1) Error:
Que::JobBuffer::push and shift#test_0001_should not deadlock:
Que::Utils::Assertions::AssertionFailed: Assertion failed!
    que/lib/que/job_buffer.rb:234:in `block in push'
    que/lib/que/job_buffer.rb:254:in `synchronize'
    que/lib/que/job_buffer.rb:254:in `sync'
    que/lib/que/job_buffer.rb:233:in `push'
    que/lib/que/job_buffer.rb:65:in `block (3 levels) in push'
    que/lib/que/job_buffer.rb:62:in `times'
    que/lib/que/job_buffer.rb:62:in `block (2 levels) in push'
    que/lib/que/job_buffer.rb:61:in `reverse_each'
    que/lib/que/job_buffer.rb:61:in `block in push'
    que/lib/que/job_buffer.rb:192:in `synchronize'
    que/lib/que/job_buffer.rb:192:in `sync'
    que/lib/que/job_buffer.rb:55:in `push'
    que/lib/que/job_buffer.spec.rb:213:in `block (5 levels) in <top (required)>'
    que/lib/que/job_buffer.spec.rb:212:in `times'
    que/lib/que/job_buffer.spec.rb:212:in `block (4 levels) in <top (required)>'
```

siegy22:
@gerrypower Can I somehow help you with this?

trammel:
We were seeing this behaviour with jobs deadlocking with hundreds of thousands in the queue. After applying this patch, the queue went through roughly 1.5 million items successfully & no locks.

jasoncodes:
Confirming this patch works for us too. This PR seems to fix a hang I can reliably reproduce by queueing empty jobs in a tight loop. Usually locks up with a locked job after only 500 to 5000 iterations on my dev machine running a Rails app in production mode. With this patch I am yet to see it lock after 250k+ iterations. 👍

aleksandrs-ledovskis:
Thanks for the patch!

It makes one of our high volume (500 txn/s via 30 _que_ workers) queues stable again.

hlascelles:
This is the PR we are waiting for before going to 1.x. Thank you to all working on it!

Is there anything else waiting on it?

siegy22:
I'd be very happy to get this merged. IMHO it's good enough to be in 1.0 as of now without tackling the issue which @ebeigarts mentioned. 

We hear from a lot of people that they are running jobs without ever having deadlocks again (see the comments above), so I think this is good enough.

@gerrypower Would you be able to add some tests? See https://github.com/que-rb/que/pull/285#issuecomment-640566777 
We'll be able to release que 1.0 in the near future then 🎉

gerrypower:
Ok - will work on adding tests so we can get this merged!

lholden:
Any chance of this getting merged?

siegy22:
@gerrypower Any updates on this?

oeoeaio:
We have taken a look at the [test](https://github.com/que-rb/que/pull/285#issuecomment-640566777) contributed by @ebeigarts above and the `Que::Utils::Assertions::AssertionFailed` issue they noticed they were seeing when running against this branch. We think this is mostly related to running `JobBuffer#push` in a tight loop, which may explain why people who have deployed this patch are not seeing it in production. Still, we suggest a possible fix in #318.

There is one other issue that we think may need addressing before we merge this PR. Specifically, moving the call to `JobBuffer#shift_job` from inside the synchronized block inside `PriorityQueue#pop` up into `JobBuffer#shift` means that there a new race condition is introduced in between [checking for job availability](https://github.com/que-rb/que/pull/285/files#diff-d359ea91b8b954fe243aedaa7237df154da7974dc8a3a2b54f2c799d4ce81c19R221) and [actually shifting the job out of the buffer](https://github.com/que-rb/que/pull/285/files#diff-d359ea91b8b954fe243aedaa7237df154da7974dc8a3a2b54f2c799d4ce81c19R78). The reason this is a problem is that it now means that `JobBuffer#shift` can potentially return `nil` (if a job that was available is no longer by the time we go to shift it off), which is something it never did before. From a brief look at the code it appears that this could result in a premature shutdown of a worker thread, because if the return value from `#shift` is falsey the worker interprets this as a signal to gracefully shutdown ([this loop](https://github.com/que-rb/que/blob/master/lib/que/worker.rb#L60) will exit, resulting in the worker thread terminating)?

By changing the test above to assert on the return value of `#shift`:

```diff
       shift_threads = concurrency.times.map do
         Thread.new do
           jobs_count_per_thread.times do
-            job_buffer.shift
+            result = job_buffer.shift
+            refute result.nil?
           end
         end
       end
```

We see the anticipated failure:
```
# Running:

#<Thread:0x00007fa766a9ceb8 /Users/rob/projects/que/lib/que/job_buffer.spec.rb:218 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	4: from /Users/rob/projects/que/lib/que/job_buffer.spec.rb:219:in `block (5 levels) in <top (required)>'
	3: from /Users/rob/projects/que/lib/que/job_buffer.spec.rb:219:in `times'
	2: from /Users/rob/projects/que/lib/que/job_buffer.spec.rb:221:in `block (6 levels) in <top (required)>'
	1: from /Users/rob/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/minitest-5.10.3/lib/minitest/assertions.rb:537:in `refute'
/Users/rob/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/minitest-5.10.3/lib/minitest/assertions.rb:139:in `assert': Expected true to not be truthy. (Minitest::Assertion)


Minitest::Assertion: Expected true to not be truthy.
/Users/rob/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/minitest-5.10.3/lib/minitest/assertions.rb:139:in `assert'
/Users/rob/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/minitest-5.10.3/lib/minitest/assertions.rb:537:in `refute'
/Users/rob/projects/que/lib/que/job_buffer.spec.rb:221:in `block (6 levels) in <top (required)>'
/Users/rob/projects/que/lib/que/job_buffer.spec.rb:219:in `times'
/Users/rob/projects/que/lib/que/job_buffer.spec.rb:219:in `block (5 levels) in <top (required)>'

F

Fabulous run in 2.607956s, 0.3834 runs/s, 11503.6450 assertions/s.

  1) Failure:
Que::JobBuffer::push and shift#test_0001_should not deadlock [/Users/rob/projects/que/lib/que/job_buffer.spec.rb:221]:
Expected true to not be truthy.

1 runs, 30001 assertions, 1 failures, 0 errors, 0 skips
rake aborted!
Command failed with status (1)
/Users/rob/.rbenv/versions/2.7.2/bin/bundle:23:in `load'
/Users/rob/.rbenv/versions/2.7.2/bin/bundle:23:in `<main>'
Tasks: TOP => spec => test
(See full trace by running task with --trace)
```

We may be wrong about this being a problem, because again, we would have expected people who have deployed this patch to have noticed this kind thread attrition. Interested to hear the experience of those who are already running this code.

rolftimmermans:
We are running the patch in production since October 4, 2021 after we had a situation where 2 out of 2 workers were deadlocked and no jobs were being processed. No issues with this patch yet.

Even if this patch would cause worker shutdowns every now and then, that's preferable over a deadlock for us, because a shutdown will cause the worker to be restarted automatically.

We process about 15-25k jobs per day.

oeoeaio:
Thanks @rolftimmermans, good to know.

My brief skim read of the code makes me think that individual worker threads could terminate, reducing the worker thread count in a given process, but given that returning `nil` from `JobBuffer#shift` is caused by a race condition between two worker threads, I don't think the last worker thread would ever exit. So I'm not sure that que process itself would ever terminate (so it could then be restarted). Again, we need to do some more investigation to confirm that 1) this is actually what happens and 2) it is a problem.

ZimbiX:
Ok, I think we're finally in a position to consolidate and merge everything related to this now. There were a number of issues discovered/caused, with fixes in various places:

- The JobBuffer can deadlock - fixed in this PR
- The waiting count assertion fail - fixed in https://github.com/que-rb/que/pull/318
- Worker thread attrition due to returning nil - fixed in 5a7e997516cae0f6ba619f89b7b7daa0a8693487
- The test from @ebeigarts in a comment above

I'm going to start rebasing & merging these all into master.

PR #292, opened 2020-09-02
Add instructions for how to start Que

xcskier56:
While this is fairly self explanatory, it took me a sec to investigate and confirm this is the proper usage. This update makes the `bundle exec que` usage clear and will make this much more beginner friendly.


siegy22:
Thank you for the contribution, I agree on your point.

ZimbiX:
Nice one

Issue #280, opened 2020-03-26
Exceptions being swallowed when running que


jpaulgs:
I'm playing around with the configuration contained within `config/environment.rb`.  However if I break that configuration I end up with the following message:

```
➜  anm git:(jobs) ✗ bundle exec que
Could not load file './config/environment.rb'
``` 

Can we update the following to log the actual error encountered?

https://github.com/que-rb/que/blob/065981f26c555ac682fa9944823324f6613e060a/bin/command_line_interface.rb#L186

ZimbiX:
s/running/starting/

PR #293, opened 2020-09-09
Add more context to error message when config files fail to load.

trammel:
Currently when a configuration/app file is required, if it fails for any reason, then a simple "Could not load file" message is given.

This might be because the required file is missing, or it could be because it contains errors.

This commit adds more context to the reason for the failure, helping debug cases where the required file contains an error.


siegy22:
Thank you for your contribution, definitely a change that is welcome 😄

ZimbiX:
Nice. This resolves #280

PR #298, opened 2020-12-09
Fix lock leak on PostgreSQL 12 and later

### LINKED ISSUE #290: Locker doesn't release locks for processed jobs under some circumstances
Hi 👋

I'm noticing an unexpected accumulation of locks by our worker processes. It seems that under some circumstances, advisory locks are not being released by the locker. Here's a query that hopefully illustrates the problem:

```sql
with
  pid_locks as
    (select pl.pid, count(*) from pg_locks pl inner join que_lockers ql on pl.pid = ql.pid group by pl.pid),
  pid_locked_jobs as
    (select pl.pid, count(*) from pg_locks pl inner join que_jobs qj on pl.objid = qj.id group by pl.pid)
select ql.pid, worker_count, queues, listening, pid_locks.count as active_locks, pid_locked_jobs.count as active_jobs
from que_lockers ql
left join pid_locks on ql.pid = pid_locks.pid
left join pid_locked_jobs on pid_locked_jobs.pid = ql.pid;
```

(Our job IDs are still in 32-bit range, so I didn't need to use `classid` in the `pid_locked_jobs` join.)

```
  pid  | worker_count |  queues   | listening | active_locks | active_jobs
-------+--------------+-----------+-----------+--------------+-------------
 15649 |           12 | {default} | t         |         4255 |           3
 19095 |           12 | {default} | t         |          673 |           4
 16646 |           12 | {default} | t         |         1148 |           6
 31188 |           12 | {default} | t         |         7909 |          10
(4 rows)
```

As you can see, at the moment we have 4 Que processes running (on Heroku, running on 4 different dynos), each with 12 workers. Our throughput is on average 400 jobs/minute, peaking occasionally at 800 jobs/minute. Normally our queue is almost empty.

However, all locker processes have hundreds/thousands of unreleased locks. I've seen the total number of locks exceed 50k and I'm suspecting it lead to `out of shared memory` errors we experienced earlier this week. I can see this problem locally as well, but I still haven't narrowed down on the simple way to reproduce it.

Does/did anybody see similar behavior? 

Any ideas on what to explore/test/log would be appreciated!

Stack details:

* Que 1.0.0.beta3 (using the ActiveJob adapter)
* Rails 5.2.4.3
* Postgresql 12
* We have a custom job middleware for reporting job metrics, but I can see the problem happening even when the job middleware disabled

jasoncodes:
Fixes #290.

PostgreSQL 12 and later do not materialise all CTEs automatically. This can result in multiple scans occurring which in turn can result in `pg_try_advisory_lock` being invoked multiple times on the same job. This results in a leak as we only call unlock once. This PR forces the CTE in the locker to be materialised which matches pre-PostgreSQL 12 behaviour.

I wasn’t able to reproduce any similar issues with the polling code but it’s possible the same issue exists there. I suspect the `lock_taken` CTE being used twice in `lock_and_update_priorities` is enough to trigger the materialisation path there.

This PR also fixes the use of an obsolete column in `pg_attrdef` which was preventing a green build on PostgreSQL 12.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         ruby_version: [2.5.x, 2.6.x, 2.7.x]
         gemfile: ["4.2", "5.2", "6.0"]
-        postgres_version: [9, 10, 11]
+        postgres_version: [9, 10, 11, 12]
         exclude:
           - { gemfile: "4.2", ruby_version: "2.7.x" }
     services:
--- a/lib/que/connection.rb
+++ b/lib/que/connection.rb
@@ -119,6 +119,10 @@ module Que
       loop { break if next_notification.nil? }
     end
 
+    def server_version
+      wrapped_connection.server_version
+    end
+
     def in_transaction?
       wrapped_connection.transaction_status != ::PG::PQTRANS_IDLE
     end
--- a/lib/que/connection.spec.rb
+++ b/lib/que/connection.spec.rb
@@ -45,6 +45,14 @@ describe Que::Connection do
         end
       end
 
+      describe "server_version" do
+        it "returns the PostgreSQL server version" do
+          assert_instance_of Integer, connection.server_version
+          assert_operator connection.server_version, :>=, 9_06_00
+          assert_operator connection.server_version, :<, 99_00_00
+        end
+      end
+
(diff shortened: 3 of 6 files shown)


siegy22:
Thank you very much! 💚

PR #302, opened 2021-01-10
Fixes for Ruby v3.

stas:
Minor fixes for Ruby v3.

### DIFF
--- a/bin/command_line_interface.rb
+++ b/bin/command_line_interface.rb
@@ -219,13 +219,7 @@ OUTPUT
 
         options[:poll_interval] = poll_interval
 
-        locker =
-          begin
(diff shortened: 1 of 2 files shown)


ZimbiX:
I think I agree with the removal of the `rescue`. But I also think this PR should be closed in favour of the more progressed https://github.com/que-rb/que/pull/303 (though it sounds like there are more significant changes required than that). Maybe propose the `rescue` change separately =)

stas:
Happy to update the PR if we want it merged :upside_down_face:

ZimbiX:
I reckon this can be closed as superseded by #319. Let's open a new PR once that goes though to remove the `rescue` =)

PR #303, opened 2021-01-21
Added support for Ruby 3.0

ollym:


morgoth:
@ollym Will you work on it further? @siegy22 can we have Ruby 3 support?

siegy22:
@morgoth I've worked on it for a little bit.. There's a huge problem: Especially with compatibility for ruby 3 with existing jobs.
The job args are stored as `[arg1, arg2, { kwarg1: foo, kwarg2: bar }]` and then passed to the run method using `*`. This won't work in ruby 3 anymore because the kwargs aren't passed as actual kwargs but rather as a positional argument..

So my attempt was to split things up into args and kwargs in the jobs table, but this requires a lot of changes to still work and backwards compatibility is pretty much impossible for existing jobs in the database.. 😢

ollym:
@morgoth @seigy22 I had come to the same conclusions which is why i stopped :joy:

Ruby 3 support will introduce breaking changes and a schema upgrade to the table so also a que major version change. Way more work than i first thought and also given i'm not really up to speed with the internals, i didn't feel comfortable doing it.

nearapogee:
While you are at it, it would be great to have the args transition to a
jsonb column instead of plain json. Sometimes I like to search for that
json.

Pardon me if you have already done that, we aren't yet on the 1.x series of
releases yet.

Cheers!

siegy22:
@nearapogee https://github.com/que-rb/que/blob/master/lib/que/migrations/4/up.sql#L11-L12 In 1.x it's already JSONB as far as I can tell. 😄

PR #314, opened 2021-09-08
Fixes for Ruby 3.0 compatibility

milgner:
Not sure whether there are other places, but these were enough to get Que running in our setup.

Afaik the double-splat operator should be compatible with all Ruby versions >= 2.0.


milgner:
Oh my... only now realized that there already are other PRs for this! Going to close this one in favour of #302 then...

PR #319, opened 2021-12-16
Support Ruby 3

### LINKED ISSUE #311: Running que fails with ruby3
The que executable seems to fail with ruby 3, due to the following line:

https://github.com/que-rb/que/blob/61e1805deb8161e5c1502e518a9a44fa464d308c/bin/command_line_interface.rb#L224

replacing `options` with `**options` resolves this.

oeoeaio:
### NOTE: still in draft

Should resolve #311 
Depends on: #334

There is still a bit of cleanup to do yet (commit history is a bit of a mess), but here is a first cut at supporting ruby 3. Builds on the work of @ollym and @siegy22. 

### Notable changes

- Database schema has changed to split job arguments into `args` and `kwargs`.
- Passing a hash literal as the last job argument to be splatted into job keyword arguments is no longer supported
- Dropped support for ruby < `2.7`
- Dropped support for rails < `6.0`

### TODO

- [x] Deal with any remaining TODOs in the code
- [x] Clean up commit history
- [x] Update readme and docs
- [x] Add a GIN index on the `kwargs` column
- [x] Update documentation for `job_options` (include searching by `.enqueue`)
- [x] Once #334 is merged:
  - [x] Rebase
  - [x] Update the version of the migration here from 5 to 6
  - [x] Start recording `job_schema_version` as 2 when enqueueing jobs:
    - [x] Bump `Que.job_schema_version` to 2
    - [x] Remove the `job_schema_version` column default
- [x] Remove the old poller index
- [x] Write upgrade guide for dealing with existing jobs without downtime (running two instances of Que: old and new)
- [x] Drop gems used to run the test suite against Rubinius
- [x] Add a NOT NULL constraint to `job_schema_version`
- [x] Rename `que_poll_idx_with_job_schema_version` (from migration 5) to `que_poll_idx` (after we drop the original)
- [x] Add a test for passing job options via ActiveJob


ZimbiX (commented):
This is looking great. Nice work =)

lib/que/job.enqueue_spec.rb
```diff
@@ -194,48 +223,54 @@ class QueueSubclassJob < QueueDefaultJob
 
     describe "priority" do
       it "should respect a default priority in a job class" do
-        assert_enqueue \
-          -> { PriorityDefaultJob.enqueue 1 },
+        assert_enqueue(
+          args: -> { PriorityDefaultJob.enqueue 1 },
```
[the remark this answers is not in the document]
oeoeaio: Yep, good call. In fact, I think I would prefer switching to making the interface consistent and always passing a lambda that enqueues a job. Passing the args and kwargs through a helper method is obscuring the true behaviour of `enqueue`.

lib/que/sequel/model.spec.rb
```diff
@@ -7,7 +7,7 @@
     require "que/sequel/model"
   end
 
-  def enqueue_job(*args)
+  ruby2_keywords def enqueue_job(*args)
```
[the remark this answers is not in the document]
oeoeaio: Yes, but I think we should change all these spec helper methods to accept a callable that makes the r queue call directly rather than passing the args and kwargs through, then we don’t need it.

lib/que/job.rb
```diff
@@ -55,17 +56,17 @@ class << self
...
+          opts = args.last
+          queue = opts.delete(:queue)
+          priority = opts.delete(:priority)
+          run_at = opts.delete(:run_at)
+          job_class = opts.delete(:job_class)
+          tags = opts.delete(:tags)
+        end
 
-        args << arg_opts if arg_opts.any?
+        args, kwargs = Que.split_out_ruby2_keywords(args)
```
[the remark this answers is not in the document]
oeoeaio: Yeah, nice.

lib/que/job.spec.rb
```diff
@@ -349,6 +383,7 @@ def run(*args)
     def execute(*args)
       TestJobClass.run(*args)
     end
+    ruby2_keywords(:execute) if respond_to?(:ruby2_keywords, true)
```
oeoeaio, line 386: I suspect Job.run needs to be a ruby2_keywords methods and this is obscuring that fact in the tests. Wondering if we can use a let to make these execute methods accept a lambda?

spec/spec_helper.rb
```diff
@@ -41,7 +41,7 @@
 
 
 # Handy constants for initializing PG connections:
-QUE_URL = ENV['DATABASE_URL'] || 'postgres://postgres:@localhost/que-test'
+QUE_URL = ENV['DATABASE_URL'] || 'postgres://localhost/que-test'
```
oeoeaio, line 44: Hm, this has been changed and then changed back on two separate occasions just within this PR. Suspect we need to get to the bottom of what the correct behaviour is. I made 0614904c67a5a982e84aaeec3491803cba04144c  because the earlier change wasn’t working for me.

ZimbiX: Whatever we do for this, we should also update this in [`tasks/safe_shutdown.rb`](https://github.com/greensync/que/blob/5c5d3bccb2ef515fcfa1c5def47bf9571ae62f3a/tasks/safe_shutdown.rb#L22)

lib/que/job.enqueue_spec.rb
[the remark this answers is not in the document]
ZimbiX: I refactored them all as lambdas, and was going to make the lambda a positional argument; but then realised it'd look better as a block, so did that instead.

lib/que/sequel/model.spec.rb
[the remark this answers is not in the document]
ZimbiX: I started down that path, but it ended up making the tests look a lot more complex. The usage of `enqueue_job` is relatively simple and we control it, so it seems ok to have a helper to me. I did notice that the model specs for Sequel and ActiveRecord had diverged, so I made them consistent.

lib/que/job.rb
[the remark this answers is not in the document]
ZimbiX: Done

spec/spec_helper.rb
ZimbiX: I've rebased to remove all changes to this now that we have https://github.com/que-rb/que/pull/324

Gemfile
```diff
@@ -28,8 +28,7 @@ group :test do
 end
 
 platforms :rbx do
-  # TODO: What are the effects of not having this?
-  # gem 'rubysl', '~> 2.0'
+  gem 'rubysl', '~> 2.0'
```
ZimbiX, line 31: @maddymarkovitz What ended up being the issue with this?

keithlayne:
Sorry if I've missed this elsewhere, but what's the target for this? When it's good to go, target v2?

Gemfile
maddymarkovitz: I'm really not sure. Early on in the process @oeoeaio and I found that you had to remove the `rubysl` dependency to get the gems to install. I don't understand why, because we weren't targeting the `rbx` (Rubinius?) platform.
A separate question is why this is in here at all if Rubininus isn't officially supported.

maddymarkovitz:
@keithlayne wrote:
> Sorry if I've missed this elsewhere, but what's the target for this? When it's good to go, target v2?

I think that's the plan. We'd like to release 1.0 after we're satisfied that the current beta release doesn't have any issues, then these changes would come in as 2.0 (after another beta release in between to test them out).

Is Ruby 3 support something you need urgently or are able to test? We want to be guided by the community in terms of developing and documenting an upgrade process.

keithlayne:
Thanks for the info. It's not urgent for us, but we're pre-release so I'd be willing to roll with the beta and help with testing and upgrade process. Just hook me up with that early-bird discount.

I _think_ that this is our only dep really holding us back from trying out ruby 3 right now.

brunoprietog:
It would be quite nice if it worked with ruby 3. A few months ago I started a project with ruby 3 and it was time to decide on the job library. Que looks very interesting, but it doesn't make me feel good to have to go back to ruby 2.7.
Thanks a lot for the work you guys are doing!

compactcode:
We have switched to this branch for an upcoming ruby `3.1` project at Zepto. Still early days but for us but I'm happy to help out with the effort/testing if needed.

ZimbiX:
Hi @compactcode, thanks for the offer. Certainly some real-world testing would be great; let us know how you go.

In terms of what's left to do development-wise, it's as @maddymarkovitz said above: we just need to work out a suitable upgrade process from Que 1 for those who can't afford to clear their jobs table. I'm thinking we need to provide the ability to run both Que 1 and Que 2 simultaneously, so that existing jobs (with keyword arguments in the `args` column rather than `kwargs`) can be treated appropriately during the transition. Then cut over to Que 2 once all the Que 1 jobs are processed. We'd need a way to determine which jobs need to be handled by which worker - maybe by adding a que version column?

ZimbiX:
Rebased to avoid conflicts

CHANGELOG.md
```diff
@@ -1,3 +1,38 @@
+### 2.0.0.beta1
+
+**Preliminary release of Ruby 3 support**
+
+**Notable changes:**
+ *   Database schema has changed to split job arguments into args and kwargs.
+ *   It's no longer possible to pass keyword arguments whose keys conflict with job option keys.
```
ZimbiX, line 7: @oeoeaio I think this point could do with clarification. With the addition of the `job_options` kwarg and deprecation of including job options straight in the kwargs [in 1.2.0](https://github.com/que-rb/que/blob/master/CHANGELOG.md#120-2022-02-23), there's no conflict anymore, right? Or do we still need to remove the deprecated code path?

ZimbiX:
I set out this afternoon planning to rebase this branch on master, and then rebase it on [the PR that adds the job schema version column](https://github.com/que-rb/que/pull/334) that we need for the upgrade process before we can release Que 2. Frustratingly, I've been in rebase hell for _hours_; finally having emerged with a new branch that's only rebased on master (with a bunch of commits squashed for my sanity): [`ruby3-redone`](https://github.com/que-rb/que/commits/ruby3-redone). I was hoping to push that to `greensync:ruby3` to update this PR, but infuriatingly, it's not passing the build. All the regular specs pass, but the ActiveJob ones do not. That's due to [the Que adapter that lives in ActiveJob still using the Que job options as top-level keys](https://github.com/rails/rails/blob/b6d4269315dcf9e8bbc8bdeec3b419df32a03154/activejob/lib/active_job/queue_adapters/que_adapter.rb#L21) rather than the new `que_options` keyword - support for which I dropped part-way through the new branch :face_with_head_bandage:

I'm not sure what to do from here; I'll have to work something out tomorrow.

ZimbiX:
I reordered the commits on [`ruby3-redone`](https://github.com/que-rb/que/compare/master...ruby3-redone) to slot the ActiveJob adapter patch in just before the commit where I dropped support for top-level job options, and triggered a build for each commit - it's now green all the way along, finally! :tada: I didn't think we'd get there, haha :sweat_smile:

Screenshot from 2022-02-25 00-40-47

I'll now force-push this PR's branch to match that one. And rebase on https://github.com/que-rb/que/pull/334 tomorrow (which will lose all the ticks, hence the screenshot).

CHANGELOG.md
ZimbiX: Resolved - I've dropped support for top-level job options and reworded this.

lib/que/job.spec.rb
ZimbiX: Has this been resolved now? I can't find the piece of code you were pointing at anymore

oeoeaio:
> Change the column default to 2?

I think this is probably fine. The only way this could be an issue is if someone continues to run a version of their app that schedules jobs using the v1 job schema without explicitly specifying that this is the case. We were thinking of just removing the default and making it a required field so there is absolutely no ambiguity going forward.

ZimbiX:
> We were thinking of just removing the default and making it a required field so there is absolutely no ambiguity going forward.

Ah, that's a good idea! I'll add that to the migration here

ZimbiX:
> - [ ] Remove the old poller index

Any idea what this refers to?

ZimbiX:
Rebased on master, and completed the post-https://github.com/que-rb/que/pull/334 actions

ZimbiX:
For the simultaneous running of old and new workers, it look like we can run two different versions of a gem like so:

```bash
sed -i -E "s/(VERSION = ).*/\\1'2.0.0'/" ../que/lib/que/version.rb
bundle init
echo "gem 'que', '~> 2.0', path: '../que'" >> Gemfile
bundle
mkdir que-1-gemfile
cp Gemfile Gemfile.lock que-1-gemfile
sed -i -E "s=^( *gem 'que').*=\1, '~> 1.0'=g" que-1-gemfile/Gemfile
BUNDLE_GEMFILE=que-1-gemfile/Gemfile bundle install

BUNDLE_GEMFILE=que-1-gemfile/Gemfile bundle exec que --version
bundle exec que --version
```

lib/que/migrations/6/up.sql
```diff
@@ -2,6 +2,7 @@ ALTER TABLE que_jobs
   ADD COLUMN kwargs JSONB NOT NULL DEFAULT '{}';
 
 CREATE INDEX que_jobs_kwargs_gin_idx ON que_jobs USING gin (kwargs jsonb_path_ops);
+DROP INDEX que_poll_idx;
```
oeoeaio, line 5: Worth renaming the new index to this as well?

ZimbiX: Ah, nice, I didn't know you could do that

oeoeaio:
Just realised that dropping support for passing job options as regular kwargs will break the Que adapter in ActiveJob. We’ll have to come up with a solution for that before we merge this.

ZimbiX:
Is there more to do than this? https://github.com/que-rb/que/pull/319/commits/500f9a5e23f2097df00b1c6965720a8662976786

oeoeaio:
That awesome! I don’t think we have a test for passing job options via ActiveJob, so it would be nice to have some coverage there, what do you think?

fameoflight:
Hey all, 

Running into this problem when running migration for this branch 

StandardError: An error has occurred, this and all later migrations canceled:

ERROR:  column "queue" of relation "que_jobs" already exists
/Users/hemantv/workspace/openzar/openzar-api/db/migrate/20220313182410_update_que_tables.rb:3:in `up'

Caused by:
PG::DuplicateColumn: ERROR:  column "queue" of relation "que_jobs" already exists

Any ideas what might be causing this

lib/que/version.rb
```diff
@@ -4,6 +4,6 @@ module Que
   VERSION = '1.3.1'
 
   def self.job_schema_version
-    1
+    2
```
hlascelles, line 7: Can line 4 above this be changed to `2.0.0.beta1` or similar? I am writing a change for que-scheduler to work with these code changes, and it will be nice to target this version by number.

hlascelles: Needed for this corresponding change: https://github.com/hlascelles/que-scheduler/pull/328

oeoeaio: Done @hlascelles!

oeoeaio:
> Running into this problem when running migration for this branch

@fameoflight could you create a separate issue to discuss this? Providing the contents of your `20220313182410_update_que_tables.rb` migration file may be useful context.

oeoeaio:
What do you think @ZimbiX @maddymarkovitz, are we ready to release this?

ZimbiX:
> Write upgrade guide for dealing with existing jobs without downtime (running two instances of Que: old and new)

@oeoeaio I'm just going to add a couple of tips for this now, but then I reckon we're ready to release a beta =)

ZimbiX:
Actually no, there's still a couple of small unresolved threads we should sort out

lib/que/job.spec.rb
oeoeaio: Yep, this was fixed in 8f474a36b0791037f81f45734e567781eb36ca7c

Gemfile
oeoeaio: I vote we just drop this. This was [added](https://github.com/que-rb/que/pull/21) back in 2014 in `0.6.0` when the test suite was running in TravisCI.

oeoeaio:
> Actually no, there's still a couple of small unresolved threads we should sort out

@ZimbiX I've added items to the TODO at the top to cover the only three unresolved threads I can see. Anything else you can think of?

ZimbiX:
Rebased to avoid conflicts in changelog and version file

ZimbiX:
Damn, I thought I'd finally finished it, and was about to merge and release, but noticed we hadn't done this so added it to the todo to tackle later:

> - Add a test for passing job options via ActiveJob

oeoeaio:
> Damn, I thought I'd finally finished it, and was about to merge and release, but noticed we hadn't done this so added it to the todo to tackle later
> - Add a test for passing job options via ActiveJob

@ZimbiX I think we covered this in 7bcda94b4f62499fca44c0f58e99e9f6c7d1bc76

ZimbiX:
> @ZimbiX I think we covered this in [7bcda94](https://github.com/que-rb/que/commit/7bcda94b4f62499fca44c0f58e99e9f6c7d1bc76)

Ah, excellent; thanks. Looks like we're good to go then! :rocket:

PR #323, opened 2021-12-22
differentiate deadlock vs timeout in job_buffer spec

smontanari:
Minor improvement to the job_buffer test used to detect deadlock


thomasvdpol:
This change is mostly executable documentation to make the intent of the test clearer. The timeout catch is there because it'd be really confusing to debug this thing if it passed after 5 seconds because things take that long for some reason that isn't a deadlock.

ZimbiX (commented):
Good idea

PR #324, opened 2021-12-23
Add Docker for easier setup of Que development environment

ZimbiX:
Providing a Docker Compose service for running PostgreSQL, with a script for running the tests, makes it a lot easier to get started for developing Que. You no longer need to have installed & started PostgreSQL, set `DATABASE_URL` appropriately, or created a database for Que. Different versions of PostgreSQL can also be tested easily, by setting `POSTGRES_VERSION` - e.g. `auto/test-postgres-14`.

This setup follows GreenSync conventions, although `auto/test` is not used on GitHub Actions.

I've used the latest Ruby Docker image for Ruby 2.7. We can update this to Ruby 3 in https://github.com/que-rb/que/pull/319.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,46 @@
+version: "3.7"
+
+services:
+
+  dev:
+    build:
+      context: .
+      target: dev-environment
+    depends_on:
+      - db
+    volumes:
+      - .:/work
+      - ruby-2.7.5-gem-cache:/usr/local/bundle
+      - ~/.docker-rc.d/:/.docker-rc.d/:ro
+    working_dir: /work
+    entrypoint: /work/scripts/docker-entrypoint
+    command: bash
+    environment:
+      DATABASE_URL: postgres://que:que@db/que-test
+
+  db:
+    image: "postgres:${POSTGRES_VERSION-13}"
+    volumes:
+      - db-data:/var/lib/postgresql/data
+    environment:
+      POSTGRES_USER: que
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+  dev:
+    build:
+      context: .
+      target: dev-environment
+    depends_on:
+      - db
+    volumes:
+      - .:/work
+      - ruby-2.7.5-gem-cache:/usr/local/bundle
+      - ~/.docker-rc.d/:/.docker-rc.d/:ro
```
maddymarkovitz, line 14: Is `~/.docker-rc.d/` a standard directory that gets created automatically by all Docker installations? If it's not, we should probably leave this out (assuming that `docker-compose up` will fail if the directory doesn't exist).

ZimbiX: It's not a Docker standard, but a GreenSync standard. I can't remember how we came up with that exact name of `~/.docker-rc.d/.docker-bashrc`. When a path you're mounting into a conainer doesn't exist, Docker creates it as a directory. With that in mind, I thought it was relatively harmless to leave in. What do you think?

maddymarkovitz: That does sound harmless but we should at least document its usage if we're going to add it to Que.
Admittedly, most Que developers outside GS probably won't want to go to the trouble of setting up their aliases to work inside one Dockerised project, but it seems remiss not to mention it.

maddymarkovitz (approved):

PR #327, opened 2022-01-21
Documentation updates in preparation for 1.0 release

maddymarkovitz:
I'm not quite sure how we should structure this but I was thinking we close off the existing `CHANGELOG.1.0.beta.md` and start a new `CHANGELOG.1.0.md`?


ZimbiX (approved):
Yeah, probably best to include the changelog in the changelog file given the previous history's there. I reckon we should have the changelog in only one place - so replace [the one attached to the release](https://github.com/que-rb/que/releases/tag/v1.0.0.beta5) with a link to this. I updated the Markdown links to better match what's there.

> I'm not quite sure how we should structure this but I was thinking we close off the existing `CHANGELOG.1.0.beta.md` and start a new `CHANGELOG.1.0.md`?

I would probably put all the 1.x changes in one file. And maybe even merge in the 0.x changelog, to form a single `CHANGELOG.md`.

chrisnankervis (approved):

maddymarkovitz:
> I would probably put all the 1.x changes in one file. And maybe even merge in the 0.x changelog, to form a single `CHANGELOG.md`.

That was my initial thought as well but looking at how it had been done in the past, I didn't know if I should leave things as they were. Still, I prefer to have it all in one place and we can always split the files again if we need to.

PR #328, opened 2022-01-25
Add backtrace to errors

trammel:
We were having issues where jobs were failing, but the error `ArgumentError` wasn't providing much information about where the problem was occurring.

This change adds a backtrace to the error logs, and also changes the other instance logging of the backtrace to be a bit more paranoid, in that the `backtrace` method may return a `nil` or an array.

I'd suggest a squash merge of these commits, if that's OK.

### DIFF
--- a/lib/que/worker.rb
+++ b/lib/que/worker.rb
@@ -138,7 +138,6 @@ module Que
           class:   error.class.to_s,
           message: error.message,
           backtrace: (error.backtrace || []).join("\n").slice(0, 10000),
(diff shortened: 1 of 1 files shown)


ZimbiX (approved):
Good idea; I've had to open up the jobs table many times to look at the backtrace. Thanks!

oeoeaio (approved):
LGTM

maddymarkovitz (approved):

PR #334, opened 2022-02-22
Add job_schema_version column

maddymarkovitz:
This is a change best made prior to merging #319.

We've added the `job_schema_version` column to the `que_jobs` table.
This will allow us to:
- In Que 2.x, start recording the `que_version` as 2
- Build a migration process from 1.x to 2.x whereby Que workers only process jobs that were enqueued with their version of Que.
- In Que 2.x, we can delete the old poller index.

We're thinking this will require a major version bump due to the additional DB migrations but we're open to suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,36 @@
+### Unreleased
+
+**ACTION REQUIRED**
+
+This release will allow you to safely upgrade to Que 2 when it comes out, without first needing to empty your `que_jobs` table.
+
+**You will need to first update to this version, apply the Que schema migration, and deploy, before you can safely begin the process of upgrading to Que 2.**
+
+Que 2 will bring Ruby 3 support, but to do that, the job arguments in the `que_jobs` table will need to be split into two columns - repurposing the existing one for positional arguments only (`args`), and adding a new one for keyword arguments (`kwargs`). This is so that Que running in Ruby 3, when reading job arguments stored in the database, can disambiguate between keyword arguments and a last positional argument hash.
+
+The args split hasn't happened yet, but when it does, we still need to be able to successfully process all the existing queued jobs which have their keyword arguments in the `args` column still. Our solution is for you to have both Que 1 workers and Que 2 workers operating simultaneously during the upgrade, each processing only the jobs enqueued from that version. Once all the Que 1 jobs are processed, the Que 1 workers can be retired.
+
+To allow the different worker versions to tell which jobs belong to which, we've added a new column to the `que_jobs` table in this version, `job_schema_version`. Jobs enqueued with Que 1 will have a `1` here, and jobs from Que 2 will have a `2`. Que schema migration 5 will default the job schema version of all existing jobs to `1`.
+
+You will need to migrate Que to the latest Que schema version (5). For instance, on ActiveRecord and Rails 6:
+
+```ruby
+class UpdateQueTablesToVersion5 < ActiveRecord::Migration[6.0]
+  def up
+    Que.migrate!(version: 5)
+  end
+  def down
+    Que.migrate!(version: 4)
+  end
+end
+```
+
+You must apply the schema migration and deploy to upgrade all workers.
+
+No further action is required from you at this stage. The Que 2 release changelog will provide full upgrade instructions for the process briefly described above of simultaneously running both Que 1 & 2 workers. Note that you won't be required to upgrade from Ruby 2.7 to Ruby 3 at the same time as upgrading to Que 2.
+
+If you use any Que plugins or custom code that interacts with the `que_jobs` table, before you can upgrade to Que 2, you will need to make sure they are updated for compatibility with Que 2: They'll need to make use of the `kwargs` column, and when inserting jobs, put the result of `Que.job_schema_version` into the `job_schema_version` column rather than continue to rely on its default of `1`.
+
 ### 1.2.0 (2022-02-23)
 
 - **Deprecated**
--- a/lib/que/locker.rb
+++ b/lib/que/locker.rb
@@ -279,7 +279,7 @@ module Que
         CURRENT_HOSTNAME,
         !!@listener,
         "{\"#{@queue_names.join('","')}\"}",
-        Que.major_version,
+        Que.job_schema_version,
       ]
     end
 
--- a/lib/que/locker.rb
+++ b/lib/que/locker.rb
@@ -279,7 +279,7 @@ module Que
         CURRENT_HOSTNAME,
         !!@listener,
         "{\"#{@queue_names.join('","')}\"}",
-        Gem::Version.new(Que::VERSION).segments.first,
+        Que.major_version,
       ]
     end
 
--- a/lib/que/locker.rb
+++ b/lib/que/locker.rb
@@ -24,8 +24,8 @@ module Que
 
   SQL[:register_locker] =
     %{
-      INSERT INTO public.que_lockers (pid, worker_count, worker_priorities, ruby_pid, ruby_hostname, listening, queues)
-      VALUES (pg_backend_pid(), $1::integer, $2::integer[], $3::integer, $4::text, $5::boolean, $6::text[])
+      INSERT INTO public.que_lockers (pid, worker_count, worker_priorities, ruby_pid, ruby_hostname, listening, queues, que_version)
+      VALUES (pg_backend_pid(), $1::integer, $2::integer[], $3::integer, $4::text, $5::boolean, $6::text[], $7::integer)
     }
 
   class Locker
@@ -279,6 +279,7 @@ module Que
         CURRENT_HOSTNAME,
         !!@listener,
         "{\"#{@queue_names.join('","')}\"}",
+        Gem::Version.new(Que::VERSION).segments.first,
       ]
     end
 
--- a/lib/que/locker.rb
+++ b/lib/que/locker.rb
@@ -24,7 +24,7 @@ module Que
 
   SQL[:register_locker] =
     %{
-      INSERT INTO public.que_lockers (pid, worker_count, worker_priorities, ruby_pid, ruby_hostname, listening, queues, que_version)
+      INSERT INTO public.que_lockers (pid, worker_count, worker_priorities, ruby_pid, ruby_hostname, listening, queues, job_schema_version)
       VALUES (pg_backend_pid(), $1::integer, $2::integer[], $3::integer, $4::text, $5::boolean, $6::text[], $7::integer)
     }
 
--- a/lib/que/migrations/5/up.sql
+++ b/lib/que/migrations/5/up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE que_jobs
+  ADD COLUMN que_version INTEGER DEFAULT 1;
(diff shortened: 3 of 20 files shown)


ZimbiX:
> migration process from 1.x to 2.x

Don't you mean 2 to 3?

lib/que/migrations/5/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE que_jobs
+  ADD COLUMN que_version INTEGER DEFAULT 1;
+CREATE INDEX que_poll_idx_with_que_version ON que_jobs (que_version, queue, priority, run_at, id) WHERE (finished_at IS NULL AND expired_at IS NULL);
```
ZimbiX, line 3: How about we drop the versionless index, `que_poll_idx`? Seems like it'd be obsolete if we're going to include the version in queries.

ZimbiX (commented):
LGTM from a quick read. I'll have a look at whether we need to check the version in any other queries

lib/que/migrations/5/up.sql
oeoeaio: Yep, it is, but we want any old (version-ignorant) pollers to keep working between running this migration and deploying the app, so we'll remove the old index and rename this one in a subsequent migration.

ZimbiX (changes requested):
Further changes are needed. Polling mode works (any v2 jobs enqueued prior to starting que remain un-run), but notify mode still runs a job with the wrong que_version.

## Test

With `app.rb` of:

```ruby
require 'bundler/setup'

require 'sequel'
require 'que'

DB = Sequel.connect(ENV['DATABASE_URL'])
Que.connection = DB

Que.migrate!(version: 5)

class MyJob < Que::Job
  def run
    puts 'Running MyJob'
  end
end
```

Start Que:

```bash
./auto/dev bash -c 'RUBYOPT= bundle exec ./bin/que ./app.rb'
```

Then enqueue a Que v2 job with SQL:

```bash
./auto/psql -c "insert into que_jobs (job_class, que_version) values ('MyJob', 2);"
```

Result:

```
Running MyJob
```

ZimbiX:
We should add specs for the behaviour too

lib/que/migrations/5/up.sql
ZimbiX: Ah, fair enough

oeoeaio:
> Polling mode works (any v2 jobs enqueued prior to starting que remain un-run), but notify mode still runs a job with the wrong que_version.

Nice @ZimbiX, yes we will need some more thorough tests for this.

ZimbiX:
I've added a failing spec.

I haven't had an easy time with the implementation. I see there as being a couple of options for addressing the notify ingestion:

1. Make the `que_job_notify` PostgreSQL function select a worker which has a matching Que version

    Pros:
    
    - Efficient
    
    Cons:
    
    - The `que_job_notify` function is already complex, and we may introduce bugs
    - May require recording the Que version elsewhere in the database, e.g. the `que_lockers` table, to be able to tell which workers are suitable

2. After the worker is notified of the new jobs, skip jobs which have the wrong Que version

    Pros:
    
    - Maybe simpler implementation

    Cons:
    
    - It's less efficient for a worker to be notified of a job, but then ignore it, and leave it to have to be picked up by a suitable worker doing a poll

    There doesn't seem to be an obvious place to put this. Maybe in the [`job_available` message resolver](https://github.com/que-rb/que/blob/f237a643b12f186e4850926d76c4295592224b37/lib/que/locker.rb#L37)? But the [`job_available` message format](https://github.com/que-rb/que/blob/f237a643b12f186e4850926d76c4295592224b37/lib/que/locker.rb#L10-L16) (what's sent to the worker by the db function) only contains `queue`, `id`, `run_at`, and `priority`. Annoyingly, the job's other columns are only loaded after the job gets locked.
    
    Approaches:

    - Add another query to load the `que_version` column early
    - Change the `job_available` message format to include `que_version`. But that necessitates a `que_job_notify` function change. And may not be backwards-compatible

4. Have multiple copies of the `que_job_notify` function - one for each Que version

    Pros:

    - Probably simpler than making the function smarter

    Cons:

    - Code duplication
    - Would resulting in another migration to deduplicate the function after the upgrade is done

3. Get users to disable notify during the upgrade process

    Pros:
    
    - No more implementation to do

    Cons:

    - Relying solely on polling will cripple throughput
    - Some users may not be aware that they need to do this, and could end up with bad data

ZimbiX:
I thought I'd first have a go at updating `que_job_notify`, since it seemed like that'd produce the cleanest result. And it was actually a lot easier than I expected!

Now the `que_version` population and checks are dynamically based off the `Que::VERSION` major version. I thought this was preferable to needing to remember and update all the places we'd put the magic number `1` for when we release Que 2 soon. If we were to, at some point, release a later major version, this would cause potentially unnecessary isolation - but who knows what changes that version would be bringing, so I think that's fine - we can change it back at that point if we want to. Or perhaps this is worth documenting above the version number constant, so you see it when updating the version?

It passes the specs I added. Manually tested with poll, notify, and threads - working great! :tada:

Screenshot from 2022-02-23 22-25-01

Note that when migrating while an old worker is running, unfortunately the following error occurs in the worker:

```
#<Thread:0x000055eb45e09eb8 /home/brendan/Projects/que/lib/que/locker.rb:161 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	8: from /home/brendan/Projects/que/lib/que/locker.rb:195:in `block in initialize'
	7: from /home/brendan/Projects/que/lib/que/locker.rb:243:in `work_loop'
	6: from /home/brendan/Projects/que/lib/que/locker.rb:275:in `cycle'
	5: from /home/brendan/Projects/que/lib/que/locker.rb:315:in `poll'
	4: from /home/brendan/Projects/que/lib/que/locker.rb:315:in `each'
	3: from /home/brendan/Projects/que/lib/que/locker.rb:324:in `block in poll'
	2: from /home/brendan/Projects/que/lib/que/poller.rb:150:in `poll'
	1: from /home/brendan/Projects/que/lib/que/connection.rb:98:in `execute_prepared'
/home/brendan/Projects/que/lib/que/connection.rb:98:in `exec_prepared': ERROR:  cached plan must not change result type (PG::FeatureNotSupported)
```

I suspect this is unavoidable, and we should just advise that if you want to avoid the error, to briefly shut down the workers while the migration is taking place. Otherwise, it should be fine once restarted by the supervisor (e.g. Kubernetes).

ZimbiX:
Commands used in testing:

```bash
# Env for all terminals:

export DATABASE_URL=postgres://que:que@localhost:5697/que-test
export PGHOST=localhost PGPORT=5697 PGUSER=que PGPASSWORD=que PGDATABASE=que-test

# Setup:

docker ps -aq | xargs -r docker stop && docker container ls -aq | xargs -r docker rm && docker volume rm -f que_db-data && docker-compose up -d db # reset db

ruby -r bundler/setup -r que -r sequel -e "Que.connection = Sequel.connect(ENV['DATABASE_URL']); Que.migrate!(version: 5)"

# Terminal 1:

bundle exec ./bin/que --min-buffer-size=1 --max-buffer-size=1 --worker-count=2 --poll-interval=1 ./app.rb

# Terminal 2:

# Update the version in `lib/que/version.rb` to `2.0.0`, then start a second worker:
bundle exec ./bin/que --min-buffer-size=1 --max-buffer-size=1 --worker-count=2 --poll-interval=1 ./app.rb

# Terminal 3:

psql -c "insert into que_jobs (job_class, que_version) values ('MyJobForQue1', 1), ('MyJobForQue1', 1), ('MyJobForQue1', 1), ('MyJobForQue1', 1), ('MyJobForQue2', 2), ('MyJobForQue2', 2), ('MyJobForQue2', 2), ('MyJobForQue2', 2)"
```

`app.rb`:

```ruby
require 'bundler/setup'

require 'sequel'
require 'que'

DB = Sequel.connect(ENV['DATABASE_URL'])
Que.connection = DB

class MyJobForQue1 < Que::Job
  def run
    puts 'Running MyJobForQue1'
  end
end

class MyJobForQue2 < Que::Job
  def run
    puts 'Running MyJobForQue2'
  end
end
```

ZimbiX (approved):

lib/que/locker.rb
```diff
@@ -266,6 +266,7 @@ def startup
         CURRENT_HOSTNAME,
         !!@listener,
         "{\"#{@queue_names.join('","')}\"}",
+        Que.major_version,
```
thomasvdpol, line 269: I really like this change but would not couple the data version to the Que major version. Sure, if you change the data version you'd want to bump the major version but the reverse might not be true - it's not inconceivable to want to bump the major version without changing the data version. I'm very much in favour of having it as a defined constant that can be easily referenced, I think `lib/que/version.rb` is a good place for it (though you could create a `lib/que/data_version.rb` or similar) but I think conflating the two versions now might cause trouble down the road.

oeoeaio: Yeah, I have been feeling more and more uncomfortable about this coupling too. I think it's an easy change to make.

ZimbiX: Ok. What name would you suggest? We should probably rename the `que_version` columns to match.

thomasvdpol: `data_schema_version` or something along those lines.

oeoeaio: It's tricky calling this a schema version because it suggests that it's something that can/ought to be considered every time there is a schema change. It's really much more limited that than in that it's only useful for additive changes to the `que_jobs` table.

oeoeaio: `job_schema_version` feels closer?

ZimbiX:
Rebased to resolve a merge conflict around the startup message.

oeoeaio (approved):
Yep, looks good, let's do it.

CHANGELOG.md
```diff
@@ -1,3 +1,36 @@
...
+
+The args split hasn't happened yet, but when it does, we still need to be able to successfully process all the existing queued jobs which have their keyword arguments in the `args` column still. Our solution is for you to have both Que 1 workers and Que 2 workers operating simultaneously during the upgrade, each processing only the jobs enqueued from that version. Once all the Que 1 jobs are processed, the Que 1 workers can be retired.
+
+To allow the different worker versions to tell which jobs belong to which, we've added a new column to the `que_jobs` table table in this version, `job_schema_version`. Jobs enqueued with Que 1 will have a `1` here, and jobs from Que 2 will have a `2`. Que schema migration 5 will default the job schema version of all existing jobs to `1`.
```
oeoeaio, line 13: table table

CHANGELOG.md
```diff
@@ -1,3 +1,36 @@
...
+  end
+end
+```
+
+You must apply the schema migration and deploy to upgrade all workers.
+
+No further action is required from you at this stage. The Que 2 release changelog will provide full upgrade instructions for the process briefly described above of simultaneously running both Que 1 & 2 workers. Note that you won't be required to upgrade from Ruby 2.7 to Ruby 3 at the same time as upgrading to Que 2.
+
+If you use any Que plugins or custom code that interacts with the `que_jobs` table, before you can upgrade to Que 2, you will need to make sure they are updated for compatibility with Que 2: They'll need to make use of the `kwargs` column, and when inserting jobs, put the result of `Que.job_schema_version` into the `job_schema_version` column rather than continue to rely on its default of `1`.
+
```
oeoeaio, line 33: Yeah, this is great.

ZimbiX:
Note: I've changed the advice around 3rd-party code using `Que.job_schema_version` in https://github.com/que-rb/que/commit/18f9479d239205a167ddb0292453443ac835609b

PR #336, opened 2022-02-23
Deprecate passing job options as top-level keyword arguments to Job.enqueue

oeoeaio:
In preparation for supporting ruby 3 keyword arguments we're deprecating support for passing in job options as top level keyword arguments to `Job.enqueue`. Instead, a dedicated `job_options` keyword will be required for passing in any job options in the future, and support for this is added by this PR.

### DIFF
--- a/lib/que/job.enqueue_spec.rb
+++ b/lib/que/job.enqueue_spec.rb
@@ -58,7 +58,7 @@ describe Que::Job, '.enqueue' do
 
   it "should be able to queue a job with complex arguments" do
     assert_enqueue [
-      1, 
+      1,
       'two',
       {
         string: "string",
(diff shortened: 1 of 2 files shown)


jpaulgs (approved):

lib/que/job.enqueue_spec.rb
```diff
@@ -122,6 +122,21 @@ def assert_enqueue(
...
+      expected_priority: 15
+  end
+
+  it "should fall back to using job options specified at the top level if not specified in job_options" do
+    assert_enqueue \
+      [1, {string: "string", run_at: Time.now + 60, priority: 10, job_options: { priority: 15 }}],
+      expected_args: [1, {string: "string", priority: 10}],
+      expected_run_at: Time.now + 60,
+      expected_priority: 15
+  end
```
jpaulgs, line 138: The title and the expectation here don't quite match.

You are passing the `priority` in via the both the `job_options` and keyword args.  I would have expected the priority to be missing from `job_options` and the `expected_priority` in the assertion to be 10.

jpaulgs: ahh it is the `run_at` option that is being tested here.  Sorry the priority appearing in both places got me messed up

ZimbiX (commented):
Nice one

Issue #333, opened 2022-02-14
Support bulk job creation


ZimbiX:
We have a need to efficiently enqueue several thousand jobs.

> The idea is, instead of opening a new transaction, enqueueing a job, and closing the transaction, you could enqueue a batch of jobs where only one argument varies between the individual jobs, and do so in a single transaction. This might provide a huge performance boost for some of our apps.
>
> -@maddymarkovitz

PR #340, opened 2022-02-25
Support bulk enqueuing of jobs which differ only in args/kwargs

oeoeaio:
Resolves: https://github.com/que-rb/que/issues/333

**Todo**:

- [x] This was built on top of the ruby3 branch (#319) - clean it up after that is merged
- [x] Complete specs for `.bulk_enqueue`
- [x] Rework interface to use block form instead with normal calls to `.enqueue`
- [x] Disable `que_job_notify` trigger in `.bulk_enqueue` for performance
- [x] Disable `que_state_notify` trigger in `.bulk_enqueue` for performance / ~~completely remove it~~ ([removal deferred](https://github.com/que-rb/que/issues/372))
- [x] Deprecate `que_state_notify` trigger in the changelog
- [x] Add option to enable notify in `.bulk_enqueue`
- [x] Test `.bulk_enqueue` in synchronous mode
- [x] Write documentation
- [x] Update mass-job-deletion docs to use `que.skip_notify`
- [x] Determine performance improvement
- [x] ~~Add a test for enqueueing a job with ActiveJob within a `.bulk_enqueue` block~~
  - ActiveJob does not have a bulk enqueue interface and ActiveJob always adds job options to the Que.enqueue call which we do not support.
- [x] Add migration instructions to changelog
- [x] ~~Confirm that the alternate instructions in the ActiveJob error message actually work for ActiveJob jobs~~ - it didn't: removed
- [x] Note ActiveJob incompatibility in the docs with alternate instructions

**To PR and release subsequently**:

- Support enqueueing different job classes within a `.bulk_enqueue` block
- Support enqueueing different job_options within a `.bulk_enqueue` block
- Deduplicate common code in enqueue and bulk enqueue paths

### DIFF
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -108,7 +108,7 @@ module Que
 
         if Thread.current[:que_jobs_to_bulk_insert]
           if self.name == 'ActiveJob::QueueAdapters::QueAdapter::JobWrapper'
-            raise Que::Error, "Que.bulk_enqueue does not support ActiveJob. Instead use: Que.enqueue(job_options: { job_class: 'MyJob' })"
+            raise Que::Error, "Que.bulk_enqueue does not support ActiveJob."
           end
 
           raise Que::Error, "When using .bulk_enqueue, job_options must be passed to that method rather than .enqueue" unless job_options == {}
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -113,15 +113,15 @@ module Que
           _run_attrs(attrs)
         else
           values =
-            if Thread.current[:jobs_to_bulk_insert].nil?
+            if Thread.current[:que_jobs_to_bulk_insert].nil?
               Que.execute(
                 :insert_job,
                 attrs.values_at(:queue, :priority, :run_at, :job_class, :args, :kwargs, :data),
               ).first
             else
               raise Que::Error, "When using .bulk_enqueue, job_options must be passed to that method rather than .enqueue" unless job_options == {}
-              Thread.current[:jobs_to_bulk_insert][:job_options]
-              Thread.current[:jobs_to_bulk_insert][:jobs_attrs] << attrs
+              Thread.current[:que_jobs_to_bulk_insert][:job_options]
+              Thread.current[:que_jobs_to_bulk_insert][:jobs_attrs] << attrs
               {}
             end
           new(values)
@@ -130,11 +130,11 @@ module Que
       ruby2_keywords(:enqueue) if respond_to?(:ruby2_keywords, true)
 
       def bulk_enqueue(job_options: {})
-        raise Que::Error, "Can't nest .bulk_enqueue" unless Thread.current[:jobs_to_bulk_insert].nil?
-        Thread.current[:jobs_to_bulk_insert] = { jobs_attrs: [], job_options: job_options }
+        raise Que::Error, "Can't nest .bulk_enqueue" unless Thread.current[:que_jobs_to_bulk_insert].nil?
+        Thread.current[:que_jobs_to_bulk_insert] = { jobs_attrs: [], job_options: job_options }
         yield
-        jobs_attrs = Thread.current[:jobs_to_bulk_insert][:jobs_attrs]
-        job_options = Thread.current[:jobs_to_bulk_insert][:job_options]
+        jobs_attrs = Thread.current[:que_jobs_to_bulk_insert][:jobs_attrs]
+        job_options = Thread.current[:que_jobs_to_bulk_insert][:job_options]
         raise Que::Error, "When using .bulk_enqueue, all jobs enqueued must be of the same job class" unless jobs_attrs.map { |attrs| attrs[:job_class] }.uniq.one?
         args_and_kwargs_array = jobs_attrs.map do |attrs|
           {
@@ -145,7 +145,7 @@ module Que
         klass = job_options[:job_class] ? Que::Job : Que.constantize(jobs_attrs.first[:job_class])
         klass._bulk_enqueue_insert(args_and_kwargs_array, job_options: job_options)
       ensure
-        Thread.current[:jobs_to_bulk_insert] = nil
+        Thread.current[:que_jobs_to_bulk_insert] = nil
       end
 
       def _bulk_enqueue_insert(args_and_kwargs_array, job_options: {})
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -107,7 +107,12 @@ module Que
         }
 
         if Thread.current[:que_jobs_to_bulk_insert]
+          if self.name == 'ActiveJob::QueueAdapters::QueAdapter::JobWrapper'
+            raise Que::Error, "Que.bulk_enqueue does not support ActiveJob. Instead use: Que.enqueue(job_options: { job_class: 'MyJob' })"
+          end
+
           raise Que::Error, "When using .bulk_enqueue, job_options must be passed to that method rather than .enqueue" unless job_options == {}
+
           Thread.current[:que_jobs_to_bulk_insert][:jobs_attrs] << attrs
           new({})
         elsif attrs[:run_at].nil? && resolve_que_setting(:run_synchronously)
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -129,7 +129,7 @@ module Que
       end
       ruby2_keywords(:enqueue) if respond_to?(:ruby2_keywords, true)
 
-      def bulk_enqueue(job_options: {})
+      def bulk_enqueue(job_options: {}, notify: false)
         raise Que::Error, "Can't nest .bulk_enqueue" unless Thread.current[:que_jobs_to_bulk_insert].nil?
         Thread.current[:que_jobs_to_bulk_insert] = { jobs_attrs: [], job_options: job_options }
         yield
@@ -143,12 +143,12 @@ module Que
           }
         end
         klass = job_options[:job_class] ? Que::Job : Que.constantize(jobs_attrs.first[:job_class])
-        klass._bulk_enqueue_insert(args_and_kwargs_array, job_options: job_options)
+        klass._bulk_enqueue_insert(args_and_kwargs_array, job_options: job_options, notify: notify)
       ensure
         Thread.current[:que_jobs_to_bulk_insert] = nil
       end
 
-      def _bulk_enqueue_insert(args_and_kwargs_array, job_options: {})
+      def _bulk_enqueue_insert(args_and_kwargs_array, job_options: {}, notify:)
         raise 'Unexpected bulk args format' if !args_and_kwargs_array.is_a?(Array) || !args_and_kwargs_array.all? { |a| a.is_a?(Hash) }
 
         if job_options[:tags]
@@ -189,7 +189,7 @@ module Que
         else
           values_array =
             Que.transaction do
-              Que.execute('SET LOCAL que.skip_notify TO true')
+              Que.execute('SET LOCAL que.skip_notify TO true') unless notify
               Que.execute(
                 :bulk_insert_jobs,
                 attrs.values_at(:queue, :priority, :run_at, :job_class, :args, :data),
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -120,7 +120,6 @@ module Que
               ).first
             else
               raise Que::Error, "When using .bulk_enqueue, job_options must be passed to that method rather than .enqueue" unless job_options == {}
-              Thread.current[:que_jobs_to_bulk_insert][:job_options]
               Thread.current[:que_jobs_to_bulk_insert][:jobs_attrs] << attrs
               {}
             end
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -162,10 +162,13 @@ module Que
           _run_attrs(attrs)
         else
           values_array =
-            Que.execute(
-              :bulk_insert_jobs,
-              attrs.values_at(:queue, :priority, :run_at, :job_class, :args, :data),
-            )
+            Que.transaction do
+              Que.execute('SET LOCAL que.skip_notify TO true')
+              Que.execute(
+                :bulk_insert_jobs,
+                attrs.values_at(:queue, :priority, :run_at, :job_class, :args, :data),
+              )
+            end
           values_array.map(&method(:new))
         end
       end
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -106,23 +106,20 @@ module Que
               raise(Error, "Can't enqueue an anonymous subclass of Que::Job"),
         }
 
-        if attrs[:run_at].nil? && resolve_que_setting(:run_synchronously)
+        if Thread.current[:que_jobs_to_bulk_insert]
+          raise Que::Error, "When using .bulk_enqueue, job_options must be passed to that method rather than .enqueue" unless job_options == {}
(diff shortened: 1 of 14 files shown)


lib/que/job.rb
```diff
@@ -101,6 +120,47 @@ def enqueue(*args)
...
+          args:     Que.serialize_json(args_and_kwargs_array),
+          data:     job_options[:tags] ? Que.serialize_json(tags: job_options[:tags]) : "{}",
+          job_class: \
+            job_options[:job_class] || name ||
+              raise(Error, "Can't enqueue an anonymous subclass of Que::Job"),
+        }
+
+        if attrs[:run_at].nil? && resolve_que_setting(:run_synchronously)
+          attrs[:args] = Que.deserialize_json(attrs[:args])
+          attrs[:kwargs] = Que.deserialize_json(attrs[:kwargs])
```
ZimbiX, line 151: I don't think this line is relevant - `kwargs` is not used in this method, as everything is sent through to the query in `args`. Is synchronous mode even tested? =S

lib/que/job.bulk_enqueue_spec.rb
```diff
@@ -0,0 +1,383 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Que::Job, '.bulk_enqueue' do
```
ZimbiX, line 5: I'd like to see a few more test cases, e.g.:

- When there are no args
- When there are no kwargs
- With various job options
- With synchronous mode

lib/que/job.rb
```diff
@@ -101,6 +120,47 @@ def enqueue(*args)
       end
       ruby2_keywords(:enqueue) if respond_to?(:ruby2_keywords, true)
 
+      def bulk_enqueue(args_and_kwargs_array, job_options: {})
```
ZimbiX, line 123: It'd probably be nice to share as much code as we can between this and the original `.enqueue` method

ZimbiX (changes requested):
This is great work.

How about sharing your manual testing runtime results here?

lib/que/job.rb
oeoeaio: Yeah, hadn’t got to updating this bit yet. I reckon it will be easy to support so we may as well.

lib/que/job.bulk_enqueue_spec.rb
oeoeaio: Yep, will do heaps more, not ready for full review yet.

lib/que/job.rb
oeoeaio: Should be easy to share the job options construction, args are treated pretty differently so we’ll have to keep that on the respective methods.

lib/que/job.bulk_enqueue_spec.rb
ZimbiX: Yep, just thought I'd note my thoughts ;)

ZimbiX:
How about making the method take an option for whether to disable the notify while enqueueing the jobs?

oeoeaio:
> How about making the method take an option for whether to disable the notify while enqueueing the jobs?

Interesting idea. How would we do that without also impacting jobs enqueued using the normal `enqueue` method? The trigger is defined on the table...

jasoncodes:
Note that disabling/enabling triggers is a schema change, meaning it will require a `ShareRowExclusiveLock`. This will prevent row modifications while the transaction is open.

I’ve found adding a trigger condition which checks a custom setting works well to temporarily stop row triggers on bulk operations. Something like this:

```sql
CREATE TRIGGER que_job_notify
AFTER INSERT ON public.que_jobs FOR EACH ROW
WHEN (NOT coalesce(current_setting('que.skip_notify', true)::boolean, false))
EXECUTE FUNCTION public.que_job_notify();
```

Then use `SET LOCAL` within the transaction to skip the trigger:

```
BEGIN;
SET LOCAL que.skip_notify TO TRUE;
…
COMMIT;
```

ZimbiX:
Rebased (via cherry-picking just the latest commit) on master now that the Ruby 3 PR has been merged

ZimbiX:
@jasoncodes We've had to adjust the query you provided, as we found it unfortunately errors after a transaction which sets the variable. It seems that `current_setting` returns empty string rather than null then, which breaks the casting to boolean:

```
➜ ./auto/psql
[+] Running 1/0
 ⠿ Container que-db-1  Running                                                                                                                                                                                                                                                 0.0s
psql (13.7 (Debian 13.7-1.pgdg110+1))
Type "help" for help.

que-test=# select coalesce(current_setting('que.skip_notify', true)::boolean, false);
 coalesce 
----------
 f
(1 row)

que-test=# begin; set local que.skip_notify to true; select coalesce(current_setting('que.skip_notify', true)::boolean, false); rollback;
BEGIN
SET
 coalesce 
----------
 t
(1 row)

ROLLBACK
que-test=# select coalesce(current_setting('que.skip_notify', true)::boolean, false);
ERROR:  invalid input syntax for type boolean: ""
```

Here's a working alternative that avoids the need to cast:

```
➜ ./auto/psql
[+] Running 1/0
 ⠿ Container que-db-1  Running                                                                                                                                                                                                                                                 0.0s
psql (13.7 (Debian 13.7-1.pgdg110+1))
Type "help" for help.

que-test=# select coalesce(current_setting('que.skip_notify', true), '') = 'true';
 ?column? 
----------
 f
(1 row)

que-test=# begin; set local que.skip_notify to true; select coalesce(current_setting('que.skip_notify', true), '') = 'true'; rollback;
BEGIN
SET
 ?column? 
----------
 t
(1 row)

ROLLBACK
que-test=# select coalesce(current_setting('que.skip_notify', true), '') = 'true';
 ?column? 
----------
 f
(1 row)

que-test=# begin; set local que.skip_notify to true; select coalesce(current_setting('que.skip_notify', true), '') = 'true'; commit;
BEGIN
SET
 ?column? 
----------
 t
(1 row)

COMMIT
que-test=# select coalesce(current_setting('que.skip_notify', true), '') = 'true';
 ?column? 
----------
 f
(1 row)
```

jasoncodes:
Good catch. I can confirm this empty string behaviour on at least PG 14.x.

Here’s an alternate solution to consider: `nullif(current_setting('que.skip_notify', true), '')::boolean`

ZimbiX:
I've just realised there's a second trigger, `que_state_notify`, which would also be triggered when inserting each job record. I've briefly looked into that previously, and iirc, it wasn't actually used by anything. So we could think about removing it. Regardless, we should investigate it.

```
➜ ./auto/psql
[...]
que-test=# \d que_jobs
[...]
Triggers:
    que_job_notify AFTER INSERT ON que_jobs FOR EACH ROW WHEN (NOT COALESCE(current_setting('que.skip_notify'::text, true), ''::text) = 'true'::text) EXECUTE FUNCTION que_job_notify()
    que_state_notify AFTER INSERT OR DELETE OR UPDATE ON que_jobs FOR EACH ROW EXECUTE FUNCTION que_state_notify()
```

---

I think I'd still like to add an option in bulk enqueueing for whether to disable the notify (disabling it by default, I reckon).

---

The `bulk_enqueue` interface is rather verbose currently:

```ruby
MyJob.bulk_enqueue(
  [
    { args: ['job1_arg1'], kwargs: { job1_kwarg1: 'x' } },
    { args: ['job2_arg1'], kwargs: { job2_kwarg1: 'x' } },
  ],
)
```

It could be good to improve that here or subsequently. I was wondering about something like:

```ruby
Que::Job.with_bulk_enqueue do
  MyJob.enqueue('job1_arg1', job1_kwarg1: 'x')
  MyJob.enqueue('job2_arg1', job2_kwarg1: 'x')
end
```

jasoncodes:
> I think because casting null to boolean doesn't

Ah, true. The trigger condition must return true and I missed the `NOT` when proposing the above.

> We think we prefer […] checking the value is set to something truthy rather than just set to something.

Agreed entirely. Comparing to `'true'` as a string doesn’t feel quite right to me but is technically fine. This string comparison is what I was trying to avoid. One final version for your consideration:

```sql
WHEN (nullif(current_setting('que.skip_notify', true), '')::boolean IS DISTINCT FROM true)
```

I read this one as “run the trigger when when skip_notify is not true”.

jasoncodes:
> I think I'd still like to add an option in bulk enqueueing for whether to disable the notify (disabling it by default, I reckon).

I think disabled by default makes sense.

> The `bulk_enqueue` interface is rather verbose currently.

Agreed. Not being able to use the normal Job `enqueue` methods is less than ideal too.

> I was wondering about something like:
> 
> ```ruby
> Que::Job.with_bulk_enqueue do
>   MyJob.enqueue('job1_arg1', job1_kwarg1: 'x')
>   MyJob.enqueue('job2_arg1', job2_kwarg1: 'x')
> end
> ```

Yes, please! This would be is a much nicer interface and would allow users to adopt bulk enqueuing without a lot of changes to application code.

---

One could divert the `insert_job` statements into a thread-local array and flush that array in a single INSERT statement at the end of the `with_bulk_enqueue` block. This should reduce the need for much of the additional implementation code for this feature by re-using almost all of the existing `enqueue` code.

The main API difference I can think of is `que_attrs` won’t be available for use in the return value from `enqueue`. I’d probably return a job object with an empty hash when `enqueue` is called in bulk mode. This should ensure things like ActiveJob’s API still work correctly (but without `provider_job_id` being set).

ZimbiX:
I've reworked the interface to use basically what I proposed. In this form, it would probably be nice to support enqueueing jobs of different job classes / job_options; and doing so is probably not hard.

To keep track of everything that's left, I've just added a todo list to the first comment.

ZimbiX:
> One could divert the `insert_job` statements into a thread-local array and flush that array in a single INSERT statement at the end of the `with_bulk_enqueue` block. This should reduce the need for much of the additional implementation code for this feature by re-using almost all of the existing `enqueue` code.

Yeah, that's what I'd been thinking too 😉

> The main API difference I can think of is `que_attrs` won’t be available for use in the return value from `enqueue`. I’d probably return a job object with an empty hash when `enqueue` is called in bulk mode. This should ensure things like ActiveJob’s API still work correctly (but without `provider_job_id` being set).

Ah, I hadn't thought of that! Cheers; done. I'm not really familiar with the ActiveJob part of Que - do you reckon we should add a test for that here?

ZimbiX:
> Agreed entirely. Comparing to `'true'` as a string doesn’t feel quite right to me but is technically fine. This string comparison is what I was trying to avoid.

Yeah, but the settings are strings, so we can't really avoid string stuff afaik, having to use either string parsing or string comparison, of which I presume the performance is similar.

> One final version for your consideration:
> 
> ```sql
> WHEN (nullif(current_setting('que.skip_notify', true), '')::boolean IS DISTINCT FROM true)
> ```

Interesting alternative. I think what we went with is slightly more readable, personally, so I'll stick with that. Thanks though

ZimbiX:
Rebased to avoid changelog conflict

ZimbiX:
With `./benchmark`:

```ruby
#!/usr/bin/env ruby

require 'bundler/setup'

require 'sequel'
require 'que'

DB = Sequel.connect(ENV['DATABASE_URL'])
Que.connection = DB

class MyJob < Que::Job; end

jobs_count = ARGV[0].to_i

DB[:que_jobs].truncate
a = Time.now
DB.transaction do
  jobs_count.times { MyJob.enqueue }
end
puts "Individual enqueues: #{'%.4f' % (Time.now - a)} seconds"

DB[:que_jobs].truncate
a = Time.now
DB.transaction do
  Que.bulk_enqueue do
    jobs_count.times { MyJob.enqueue }
  end
end
puts "Bulk enqueue:        #{'%.4f' % (Time.now - a)} seconds"
```

```
➜ export DATABASE_URL=postgres://que:que@localhost:5697/que-test

➜ ./benchmark 1
Individual enqueues: 0.0060 seconds
Bulk enqueue:        0.0023 seconds

➜ ./benchmark 10
Individual enqueues: 0.0136 seconds
Bulk enqueue:        0.0047 seconds

➜ ./benchmark 100
Individual enqueues: 0.0478 seconds
Bulk enqueue:        0.0104 seconds

➜ ./benchmark 1000
Individual enqueues: 0.6871 seconds
Bulk enqueue:        0.0839 seconds

➜ ./benchmark 10000
Individual enqueues: 7.0088 seconds
Bulk enqueue:        0.7667 seconds

➜ ./benchmark 100000
Individual enqueues: 65.6369 seconds
Bulk enqueue:        7.3095 seconds
```

So, close to 10x the speed

ZimbiX:
Here's the difference skipping notify makes:

```
➜ ./benchmark 100000 # with no worker processes running
Bulk enqueue:          7.2923 seconds
Bulk enqueue (notify): 10.1594 seconds

➜ ./benchmark 100000 # with several worker processes running
Bulk enqueue:          6.9337 seconds
Bulk enqueue (notify): 12.3999 seconds
```

With workers running, it's close to 2x the speed

jpaulgs:
Thank you for recording the difference that notify makes, as that was my first thought when I saw

```
➜ ./benchmark 1
Individual enqueues: 0.0060 seconds
Bulk enqueue:        0.0023 seconds
```

ZimbiX (approved):

ZimbiX:
<details>
  <summary>My updated benchmark script</summary>
  
  ```ruby
  #!/usr/bin/env ruby
  
  require_relative 'app'
  
  def jobs_count
    ARGV[0].to_i
  end
  
  def repetitions
    5
  end
  
  def benchmark
    DB[:que_jobs].truncate
    times = []
    repetitions.times do
      DB[:que_jobs].truncate
      a = Time.now
      yield
      b = Time.now
      duration = b - a
      times << duration
      puts "#{'%.4f' % duration} seconds"
    end
    puts
    puts "#{'%.4f' % (times.sum / times.size)} seconds average"
  end
  
  def enqueue_jobs
    jobs_count.times do
      MyJob.enqueue(1, '2', 3.0, z4: '5', z6: '7', z8: { 9 => 'ten!' })
    end
  end
  
  puts '', 'Individual enqueues:'
  benchmark do
    DB.transaction do
      enqueue_jobs
    end
  end
  
  puts '', 'Bulk enqueue:'
  benchmark do
    DB.transaction do
      Que.bulk_enqueue do
        enqueue_jobs
      end
    end
  end
  
  puts '', 'Bulk enqueue (notify):'
  benchmark do
    DB.transaction do
      Que.bulk_enqueue(notify: true) do
        enqueue_jobs
      end
    end
  end
  ```
  
  ```
  ➜ time ./benchmark 10000                                             
  
  Individual enqueues:
  8.1163 seconds
  7.0399 seconds
  6.9656 seconds
  7.6106 seconds
  8.7115 seconds
  
  7.6888 seconds average
  
  Bulk enqueue:
  0.7582 seconds
  0.8517 seconds
  0.8171 seconds
  0.8359 seconds
  0.7452 seconds
  
  0.8016 seconds average
  
  Bulk enqueue (notify):
  1.3392 seconds
  1.4035 seconds
  1.4391 seconds
  1.4722 seconds
  1.6834 seconds
  
  1.4675 seconds average
  ./benchmark 10000  17.58s user 1.44s system 37% cpu 50.436 total
  ```
</details>

oeoeaio (commented):
I can't approve because it's technically my PR, but love your work @ZimbiX!

jpaulgs (approved):

ZimbiX:
I just remembered a couple of small things we forgot to do, so have updated the todo list

ZimbiX (approved):

ZimbiX:
:rocket:

Issue #345, opened 2022-03-03
Empty job buffer not refilled when any worker priority > highest job priority


oeoeaio:
### Expected behaviour
If workers of an appropriate priority are running then the locker/poller should work to push available jobs to workers as soon as they run out of jobs. This happens most of the time.

### Actual behaviour
With particular worker / job priority combinations, polling seems to only occur (or occur successfully) on the polling interval rather than when workers run out of jobs.

### Steps to reproduce
- Enqueue lots of no-op jobs with priority of 99
- Run `que` with the following config:

#### `--worker-count 2 --worker-priorities 98,99 --poll-interval 5 --maximum-buffer-size 0`
- Two jobs will be processed every 5 seconds

#### `--worker-count 2 --worker-priorities 98,99 --poll-interval 5 --maximum-buffer-size 5`
- Six jobs will be processed every 5 seconds 
- Assume `buffer_size + 1` is some quirk of how jobs are fed to workers?

#### `--worker-count 2 --worker-priorities 99,99 --poll-interval 5 --maximum-buffer-size 0`
- Jobs will be processed continuously, with the poller pushing new jobs to workers as soon as they are free

#### `--worker-count 2 --worker-priorities 99,150 --poll-interval 5 --maximum-buffer-size 0`
- Same as above, jobs will be processed continuously, with the poller pushing new jobs to workers as soon as they are free

Haven't had a chance to look and understand what is causing this yet, but just wanted to document this here so I don't forget.

ZimbiX:
I assume you meant `--worker-count`; but that's actually redundant when providing `--worker-priorities`.

And with `--maximum-buffer-size 0`, then `--minimum-buffer-size` is required. Were you using 0 for this in all cases?

ZimbiX:
Note that a worker with priority 98 is supposed to only work jobs with a priority number <= 98 ([docs](https://github.com/que-rb/que/tree/master/docs#worker-priorities-and-worker-count)). From the readme:

> We use the Linux priority scale - a lower number is more important.

ZimbiX:
Reproduced:

```
# Reset database
➜ docker ps -aq | xargs -r docker stop && docker container ls -aq | xargs -r docker rm && docker volume rm -f que_db-data && docker-compose up -d db

# Migrate database
➜ export DATABASE_URL=postgres://que:que@localhost:5697/que-test
[...]

PR #348, opened 2022-03-03
Continue polling for jobs when only some of the waiting worker priorities will be fully utilised

ZimbiX:
Resolves https://github.com/que-rb/que/issues/345.

Basically, this is all about polling again immediately if there are a decent number of jobs we need to get through. But 'decent number of jobs' needs a good definition.

Previously, all idle worker priorities needed to be fully utilised by the jobs from a poll in order to not sleep before the next poll. This change makes it so that sleeping is skipped if *any* waiting worker priority is fully satisfied by the jobs from the poll.

Determining whether a worker priority would be fully utilised turned out to be far less trivial than I first thought it'd be. You have to take into account that jobs with a higher priority (a smaller number) than a worker's priority also match it.

I'm sure there's a more functional programming way to do this, but I couldn't think of it!

This was kinda hard to explain, so wording suggestions more than welcome.

A simpler alternative would be to just sleep if there are no jobs returned from the poll. That'd incur an extra poll when only a small number of jobs come in, but might be worth the ability to avoid this PR's increase in code complexity? I've kinda had to reimplement allocating jobs to workers!

Manually tested:

```
➜ ruby -r bundler/setup -r ./app.rb -e 'DB[:que_jobs].truncate; [1, 2].each { |priority| 10.times { MyJob.enqueue(job_options: { priority: priority }) } }' && bundle exec que --worker-priorities 1,2 --poll-interval 2 --minimum-buffer-size 0 --maximum-buffer-size 0 --log-level debug --log-internals ./app.rb | grep -E '(newly_locked|Running MyJob)'
I, [2022-03-04T00:09:57.981221 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:57.981175Z","queue":"default","locked":2,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503879,805503880]}
I, [2022-03-04T00:09:57.983762 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:57.988989 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.036137 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.036092Z","queue":"default","locked":2,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503881,805503882]}
I, [2022-03-04T00:09:58.038124 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.038463 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.091545 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.091499Z","queue":"default","locked":2,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503883,805503884]}
I, [2022-03-04T00:09:58.092894 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.093421 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.147326 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.147236Z","queue":"default","locked":2,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503885,805503886]}
I, [2022-03-04T00:09:58.150473 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.150893 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.202083 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.202014Z","queue":"default","locked":2,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503887,805503888]}
I, [2022-03-04T00:09:58.205669 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.205723 #2692386]  INFO -- : Running MyJob - priority: 1
I, [2022-03-04T00:09:58.258079 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.257993Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503889]}
I, [2022-03-04T00:09:58.260309 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.313026 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.312945Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503890]}
I, [2022-03-04T00:09:58.315538 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.366900 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.366867Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503891]}
I, [2022-03-04T00:09:58.368784 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.420966 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.420890Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503892]}
I, [2022-03-04T00:09:58.423198 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.474958 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.474882Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503893]}
I, [2022-03-04T00:09:58.477672 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.529960 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.529878Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503894]}
I, [2022-03-04T00:09:58.532743 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.585029 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.584941Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503895]}
I, [2022-03-04T00:09:58.586900 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.639576 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.639492Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503896]}
I, [2022-03-04T00:09:58.642056 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.694544 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.694449Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503897]}
I, [2022-03-04T00:09:58.697122 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.749006 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.748908Z","queue":"default","locked":1,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[805503898]}
I, [2022-03-04T00:09:58.751463 #2692386]  INFO -- : Running MyJob - priority: 2
I, [2022-03-04T00:09:58.802323 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:09:58.802238Z","queue":"default","locked":0,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[]}
I, [2022-03-04T00:10:00.826135 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:10:00.826047Z","queue":"default","locked":0,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[]}
I, [2022-03-04T00:10:02.851622 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:10:02.851551Z","queue":"default","locked":0,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[]}
I, [2022-03-04T00:10:04.874819 #2692386]  INFO -- : {"lib":"que","hostname":"ZimbiX-GreenSync","pid":2692386,"thread":2360,"internal_event":"poller_polled","object_id":2420,"t":"2022-03-03T13:10:04.874719Z","queue":"default","locked":0,"priorities":{"2":1,"1":1},"held_locks":[],"newly_locked":[]}
^C
```

### DIFF
--- a/lib/que/poller.rb
+++ b/lib/que/poller.rb
@@ -269,6 +269,13 @@ module Que
     private
 
     def any_priority_satisfied?(priorities, jobs)
+      worker_jobs = allocate_jobs_to_workers(priorities, jobs)
+      priorities.any? do |worker_priority, waiting_workers_count|
+        worker_jobs[worker_priority].length == waiting_workers_count
+      end
+    end
+
+    def allocate_jobs_to_workers(priorities, jobs)
       jobs = jobs.sort_by { |job| job.fetch(:priority) }
       worker_jobs = {}
       priorities.each do |worker_priority, waiting_workers_count|
@@ -282,9 +289,7 @@ module Que
           end
         end
       end
-      priorities.any? do |worker_priority, waiting_workers_count|
-        worker_jobs[worker_priority].length == waiting_workers_count
-      end
+      worker_jobs
     end
   end
 end
--- a/lib/que/poller.rb
+++ b/lib/que/poller.rb
@@ -146,8 +146,6 @@ module Que
 
       return unless should_poll?
 
-      expected_count = priorities.inject(0){|s,(_,c)| s + c}
-
       jobs =
         connection.execute_prepared(
           :poll_jobs,
--- a/lib/que/poller.rb
+++ b/lib/que/poller.rb
@@ -274,10 +274,9 @@ module Que
       priorities.any? do |worker_priority, waiting_workers_count|
         waiting_workers_count.times do
           return false if job_priorities.empty?
-          if job_priorities.first <= worker_priority
-            job_priorities.shift
-            worker_job_counts[worker_priority] += 1
-          end
+          break if job_priorities.first > worker_priority
+          job_priorities.shift
+          worker_job_counts[worker_priority] += 1
         end
         worker_job_counts[worker_priority] == waiting_workers_count
       end
--- a/lib/que/poller.rb
+++ b/lib/que/poller.rb
@@ -269,27 +269,25 @@ module Que
     private
 
     def any_priority_satisfied?(priorities, jobs)
-      worker_jobs = allocate_jobs_to_workers(priorities, jobs)
+      worker_job_counts = workers_jobs_allocation_counts(priorities, jobs)
       priorities.any? do |worker_priority, waiting_workers_count|
-        worker_jobs[worker_priority].length == waiting_workers_count
+        worker_job_counts[worker_priority] == waiting_workers_count
       end
     end
 
-    def allocate_jobs_to_workers(priorities, jobs)
-      jobs = jobs.sort_by { |job| job.fetch(:priority) }
-      worker_jobs = {}
+    def workers_jobs_allocation_counts(priorities, jobs)
+      job_priorities = jobs.map { |job| job.fetch(:priority) }.sort
+      worker_job_counts = Hash.new(0)
       priorities.each do |worker_priority, waiting_workers_count|
-        worker_jobs[worker_priority] = []
         waiting_workers_count.times do
-          next if jobs.empty?
-          if jobs.first.fetch(:priority) <= worker_priority
-            job = jobs.shift
-            worker_jobs[worker_priority] ||= []
-            worker_jobs[worker_priority] << { id: job.fetch(:id), priority: job.fetch(:priority) }
+          next if job_priorities.empty?
+          if job_priorities.first <= worker_priority
+            job_priorities.shift
+            worker_job_counts[worker_priority] += 1
           end
         end
       end
-      worker_jobs
+      worker_job_counts
     end
   end
 end
--- a/lib/que/poller.rb
+++ b/lib/que/poller.rb
@@ -269,25 +269,18 @@ module Que
     private
 
     def any_priority_satisfied?(priorities, jobs)
-      worker_job_counts = workers_jobs_allocation_counts(priorities, jobs)
-      priorities.any? do |worker_priority, waiting_workers_count|
-        worker_job_counts[worker_priority] == waiting_workers_count
-      end
-    end
-
-    def workers_jobs_allocation_counts(priorities, jobs)
       job_priorities = jobs.map { |job| job.fetch(:priority) }.sort
       worker_job_counts = Hash.new(0)
-      priorities.each do |worker_priority, waiting_workers_count|
+      priorities.any? do |worker_priority, waiting_workers_count|
         waiting_workers_count.times do
-          next if job_priorities.empty?
+          return false if job_priorities.empty?
           if job_priorities.first <= worker_priority
             job_priorities.shift
             worker_job_counts[worker_priority] += 1
           end
         end
+        worker_job_counts[worker_priority] == waiting_workers_count
       end
-      worker_job_counts
     end
   end
 end
--- a/lib/que/poller.rb
+++ b/lib/que/poller.rb
@@ -159,7 +159,7 @@ module Que
         )
 
       @last_polled_at      = Time.now
-      @last_poll_satisfied = expected_count == jobs.count
+      @last_poll_satisfied = any_priority_satisfied?(priorities, jobs)
 
       Que.internal_log :poller_polled, self do
         {
@@ -263,6 +263,27 @@ module Que
           DROP FUNCTION pg_temp.lock_and_update_priorities(jsonb, que_jobs);
           DROP TYPE pg_temp.que_query_result;
         SQL
+      end
+    end
+
+    private
+
+    def any_priority_satisfied?(priorities, jobs)
+      jobs = jobs.sort_by { |job| job.fetch(:priority) }
+      worker_jobs = {}
+      priorities.each do |worker_priority, waiting_workers_count|
+        worker_jobs[worker_priority] = []
+        waiting_workers_count.times do
+          next if jobs.empty?
+          if jobs.first.fetch(:priority) <= worker_priority
+            job = jobs.shift
+            worker_jobs[worker_priority] ||= []
+            worker_jobs[worker_priority] << { id: job.fetch(:id), priority: job.fetch(:priority) }
+          end
+        end
+      end
+      priorities.any? do |worker_priority, waiting_workers_count|
+        worker_jobs[worker_priority].length == waiting_workers_count
       end
     end
   end
--- a/lib/que/poller.spec.rb
+++ b/lib/que/poller.spec.rb
@@ -255,29 +255,61 @@ describe Que::Poller do
       assert poller.should_poll?
     end
 
-    it "should be true if the last poll returned a full complement of jobs" do
-      jobs = 5.times.map { Que::Job.enqueue }
+    it "should be true if the jobs returned from the last poll satisfied all given priorities" do
+      job_ids_p10 = 3.times.map { Que::Job.enqueue(job_options: { priority: 10 }).que_attrs[:id] }
+      job_ids_p20 = 2.times.map { Que::Job.enqueue(job_options: { priority: 20 }).que_attrs[:id] }
 
-      result = poller.poll(priorities: {500 => 3}, held_locks: Set.new)
-      assert_equal 3, result.length
+      result = poller.poll(priorities: { 10 => 3, 20 => 2 }, held_locks: Set.new)
+      assert_equal (job_ids_p10 + job_ids_p20), result.map(&:id)
 
       assert_equal true, poller.should_poll?
     end
 
-    it "should be false if the last poll didn't return a full complement of jobs" do
-      jobs = 5.times.map { Que::Job.enqueue }
+    it "should be true if the jobs returned from the last poll satisfied any given priority" do
+      job_ids_p10 = 2.times.map { Que::Job.enqueue(job_options: { priority: 10 }).que_attrs[:id] }
+      job_ids_p20 = 2.times.map { Que::Job.enqueue(job_options: { priority: 20 }).que_attrs[:id] }
 
-      result = poller.poll(priorities: {500 => 7}, held_locks: Set.new)
-      assert_equal 5, result.length
+      result = poller.poll(priorities: { 10 => 2, 20 => 1 }, held_locks: Set.new)
+      assert_equal (job_ids_p10 + [job_ids_p20.first]), result.map(&:id)
+
+      assert_equal true, poller.should_poll?
+    end
+
+    it "should be true if the jobs returned from the last poll satisfied any given priority and were slightly higher priority than each priority requested" do
+      job_ids_p10 = 2.times.map { Que::Job.enqueue(job_options: { priority: 10 }).que_attrs[:id] }
+      job_ids_p20 = 2.times.map { Que::Job.enqueue(job_options: { priority: 20 }).que_attrs[:id] }
+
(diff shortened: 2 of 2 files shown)


lib/que/poller.rb
```diff
@@ -265,5 +263,21 @@ def cleanup(connection)
...
+      worker_job_counts = Hash.new(0)
+      priorities.any? do |worker_priority, waiting_workers_count|
+        waiting_workers_count.times do
+          return false if job_priorities.empty?
+          break if job_priorities.first > worker_priority
+          job_priorities.shift
+          worker_job_counts[worker_priority] += 1
+        end
+        worker_job_counts[worker_priority] == waiting_workers_count
+      end
```
oeoeaio, line 280: Yeah, I see how this is more complicated than it first sounded, nice work.

I want to understand the relationship between the buffer size, the `available_priorities` hash and what is returned by the poller a bit better. I'm wondering whether a better way to determine whether a poll is satisfied is if we collect at least enough jobs to fill whatever buffer space is available (and maybe at least one priority?, but that feels like an anti-optimisation given the complexity of implementing it). My understanding at the moment is that the counts in `job_buffer.available_priorities` sum to the number of available workers plus the available space in the buffer. If we can fill the buffer then I think that represents a satisfactory poll, regardless of whether some workers didn't get a job.

oeoeaio: Just had another thought, maybe all that we care about is that the lowest priority worker queue is full?

The available space in the buffer is added to the [lowest priority worker queue](https://github.com/que-rb/que/blob/master/lib/que/job_buffer.rb#L124) when constructing the `available_priorities` hash. So if the lowest priority worker queue is full then the buffer will be full and all available workers of the lowest priority will have a job to do. I think that is a pretty good metric to use for `last_poll_satisfied`.

oeoeaio: Just thinking this out:

#### Five available p1 workers, five available p50 workers, 100 p1 jobs, 3 p2 jobs, buffer size of 10
{ p1: 5, p50: 15 }, so lowest priority queue is satisfied

#### Five available p1 workers, five available p50 workers, 3 p1 jobs, 100 p2 jobs, buffer size of 10
{ p1: 3, p50: 15 }, so lowest priority queue is satisfied

#### Five available p1 workers, five available p50 workers, 10 p1 jobs, 10 p2 jobs, buffer size of 10
{ p1: 5, p50: 15 }, so lowest priority queue is satisfied

#### Five available p1 workers, five available p50 workers, 9 p1 jobs, 10 p2 jobs, buffer size of 10
{ p1: 5, p50: 14 }, so lowest priority queue is *NOT* satisfied, but I think I am OK with that.

oeoeaio: But, maybe that doesn't actually simplify anything because to determine whether the lowest priority has been satisfied we need to do exactly what you have done and allocate jobs to higher priorities.

Another thought: a more fuzzy way of doing this is to just compare the number of jobs to the lowest priority availability without being too exact about where the jobs will actually end up?

Just thinking this out:

Five available p1 workers, five available p50 workers, available buffer space of 10, the available priorities will be
`{ p1: 5, p50: 15 }`

| p1 jobs| p2 jobs | total job | total jobs > p50 availability? (i.e. last_poll_satisfied?) |
| --- | --- | --- | --- |
| 100 | 3 | 20 | yes |
| 3   | 100 | 18 | yes |
| 10 | 10 | 15 | yes |
| 9 | 10 | 19 | yes |
| 4 | 10 | 14 | no |
| 10 | 4 | 14 | no |

That looks pretty good to me. Will see if I can break it.

oeoeaio: See #349

ZimbiX: I wanted a simpler way to understand the satisfaction determination, so tried to come up with a way to visualise it; and I think [this](https://docs.google.com/drawings/d/1F6J9MB_uDMrZ0RAVxXLRkwao8Bc8pLrVEx7lbEKOGko/edit?usp=sharing) is pretty good:

Que poller_ worker priorities satisfaction (1)

The requested priorities are represented as a physical container, with sections of increasing width. The width of each section corresponds to the priority (priority 1 is thin, 2 is wider, 3 would be even wider, etc.). The height of each section corresponds to the number of that priority requested.

The jobs are represented as physical blocks, of varying width. The width of a block corresponds to the job's priority (priority 1 is thin, 2 is wider, 3 would be even wider, etc.).

Allocation happens by ordering the blocks by width and then dropping them into the container against its left edge. Blocks fall down until they can't squeeze through any further. The allocations are then which blocks sit in which section.

Determining the satisfaction (whether any requested priority is full of jobs) is done by seeing if any section is vertically full.

ZimbiX: I don't think your simplified algorithm in https://github.com/que-rb/que/pull/349 meets the satisfaction requirements. Consider the first red case in my diagram. https://github.com/que-rb/que/pull/349 would mistakenly consider that as satisfied, since the jobs count (3) is equal to the requested count of the lowest priority (p2: 3).

oeoeaio: Yeah, the algorithm in #349 is fuzzy, but I think it offers a very simple good-enough algorithm at the cost of an occasional re-poll when there may not be a huge number of jobs to work. 

I have tried pretty hard to think of a case where the fuzziness actually results in a problematic behaviour (i.e. a performance hit), and I haven't really come up with anything. Worst case scenario is when there are a lot of high priority workers, few lower priority workers, no/small buffer and a steady stream of high priority jobs that is not quite enough to saturate the high priority workers but more than enough to saturate the low priority workers. In that scenario we would continually be polling and only actually seeing some of the high priority workers actually do anything. It's a fine balance though, as soon as the stream of jobs becomes more than enough for the high priority workers to handle or slow enough for the number of available jobs each poll to drop below the number of low priority workers plus the buffer it will behave as expected.

oeoeaio: If the number of low priority workers plus the buffer size is greater than the total number of workers (as is the case with the default settings `priorities: 10,20,50,any,any,any and buffer: 8`) then it's very difficult to think of a case where you get more than one potentially unnecessary poll occasionally.

oeoeaio: Just to clarify, #349 is never intended to answer the question of `whether any requested priority is full of jobs`. It's a different question, which I think provides a close enough approximation.

oeoeaio: My main concerns with this algorithm are the ones you raised yourself: that it's complex, a bit impenetrable and that we're re-implementing worker allocation which introduces a coupling we'd probably rather not have. Happy to go with it though if you don't think #349 is appropriate.

ZimbiX: Ah, I see. I suppose for a rough algorithm, it is better that it err on the side of caution - sometimes polling again unnecessarily, rather than mistakenly sleeping.

Your code is way more readable than mine. And yeah, I can't see that there's much of a drawback either. So I'm leaning towards going with that. But I'd like to let it sit for a while to see how I feel about it later.

> if there is a steady stream of jobs one could argue that polling continuously is actually the preferred behaviour, even if workers are not being fully utilised?

I would have thought the notify would handle most/all of that case; but I'm not totally across it.

oeoeaio: > I'd like to let it sit for a while to see how I feel about it later.

Sure, sounds good.

> I would have thought the notify would handle most/all of that case; but I'm not totally across it.

Oh yeah! I completely forgot about that, you’re probably right.

ZimbiX:
Ah crap, wrong button!!

PR #349, opened 2022-03-03
Continue polling when the last poll returned more jobs that the lowest priority request

oeoeaio:
Resolves https://github.com/que-rb/que/issues/345.

Alternative to https://github.com/que-rb/que/pull/348

### DIFF
--- a/lib/que/poller.rb
+++ b/lib/que/poller.rb
@@ -146,8 +146,6 @@ module Que
 
       return unless should_poll?
 
-      expected_count = priorities.inject(0){|s,(_,c)| s + c}
-
       jobs =
         connection.execute_prepared(
           :poll_jobs,
@@ -159,7 +157,7 @@ module Que
         )
 
       @last_polled_at      = Time.now
-      @last_poll_satisfied = expected_count == jobs.count
+      @last_poll_satisfied = poll_satisfied?(priorities, jobs)
 
       Que.internal_log :poller_polled, self do
         {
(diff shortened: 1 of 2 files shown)


ZimbiX (changes requested):
See https://github.com/que-rb/que/pull/348#discussion_r819217705

lib/que/poller.rb
```diff
@@ -265,5 +263,12 @@ def cleanup(connection)
         SQL
       end
     end
+
+    private
+
+    def poll_satisfied?(priorities, jobs)
+      lowest_priority = priorities.keys.max
```
oeoeaio, line 270: This was just a first crack at this, can probably do something a little more robust to select the requests for the lowest priority workers…

ZimbiX: What's not robust about it? Looks like it's always an integer - [`any` has been replaced by nil](https://github.com/que-rb/que/blob/e51f6c1826f12ccf07018b1cee3c9adfb456b3f7/bin/command_line_interface.rb#L72-L73) and then [nil replaced by the priority maximum](https://github.com/que-rb/que/blob/e51f6c1826f12ccf07018b1cee3c9adfb456b3f7/lib/que/job_buffer.rb#L128).

oeoeaio: Oh! I missed that. It’s probably fine then.

ZimbiX (approved):
Yeah, let's go with this over https://github.com/que-rb/que/pull/348. Simple's good

PR #361, opened 2022-05-16
Move spec files under `spec`

JoeSouthan:
To conform with general best practice it's better to keep your tests under `spec` or `tests` rather than next to your code


ZimbiX (approved):
I agree that having them under `spec/` is preferable. Thanks!

Issue #325, opened 2022-01-06
Add que_jobs.initial_run_at ? 


paul110:
Hi, 

This is somewhat similar to this (quite old) issue https://github.com/que-rb/que/issues/33 (which I think would likely still be a "nice to have") but we think that an `initial_run_at` would likely be a lot more useful. 

Our use case is that we are trying to see how backed up Que jobs are in order to set alarms based on that. The issue is that currently in order to determine the `job_delay`, we need to do something along these lines: 
- if `error_count = 0` => `job_delay = now() - run_at`
- if `error_count > 0`   =>  `job_delay = now() - F(run_at, error_count)`

The major benefit in adding  `initial_run_at` would be to make this always be: 
```
job_delay = now() - initial_run_at
```
regardless of the `error_count`.

Is there any appetite for adding the column `initial_run_at` (and also potentially `created_at` as a nice to have)? 

Thanks,
Paul

hlascelles:
If we take into account (thank you @owst) the job execution time, it would be `job_delay = now() - initial_run_at - elapsed`.

Issue #267, opened 2020-01-13
[RFC] Adding Prometheus instrumentation to Que


lawrencejones:
Hey team!

I'm opening this issue to get a sense of whether Que maintainers are supportive of an attempt to instrument the upstream Que project with native Prometheus metrics. We have an old fork of Que that we've modified to include this instrumentation, but are keen to move back to upstream- before we try porting our metrics, it would be good to understand if the project would be receptive to this attempt!

I've attached a dashboard from one of our staging environments as an example of the top-level metrics we have here. We've also tracked internal processes like job lock acquisition, which has helped us alert on degraded performance (as an example, whenever long snapshots are held open) in the past.

Is this something that would be welcomed by the maintainers?

Cheers,
Lawrence

![screencapture-grafana-gocardless-io-d-GyEPq1Zmz-que-2020-01-13-11_33_18](https://user-images.githubusercontent.com/3518874/72255391-8191c200-35fe-11ea-8d42-ba4b6d885025.png)

siegy22:
I think in general we're open to such requests!

We're currently working to get a stable 1.x release: https://github.com/que-rb/que/milestone/1
So this feature will not be our number one priority.

@chanks (as the original author) what do you think about this?

chanks:
Never heard of Prometheus, maybe I've been out of the Ruby game too long. I think instrumentation in general is a good thing, but if Prometheus is a specific product (as opposed to a standard?) then I'd suggest having a more generic instrumentation layer in place that it could hook into. Maybe as part of a third-party gem, maybe not.

lawrencejones:
Ah, I'd made an incorrect assumption that Prometheus would be familiar. In 2018 it was adopted by the cloud native foundation and is by many measures the more popular metrics system in operation nowadays.

Prometheus scrapes HTTP endpoints that provide metrics in the https://openmetrics.io/ text exposition format (the Prometheus project was the driving force behind establishing the format specification). It's usually a bit difficult to put these things behind a layer, but if we get round to this then I'll give it a shot.

pirj:
@lawrencejones would it be possible to plug it, and put the functionality in a separate gem?

hlascelles:
Linking this to an issue which is about worker health checks (but where we have touched on Prometheus) https://github.com/que-rb/que/issues/260

We would be extremely interested in using this, let us know if you need help @lawrencejones 
[...]

PR #362, opened 2022-05-17
Expose job latency metric via ActiveSupport Notifications job middleware

stephenbinns:
This adds ActiveSupport notifications as Job Middleware, there is also a change to expose the latency for the job to be picked up which is useful to work out how close to capacity the workers are.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -798,6 +798,32 @@ Que.job_middleware.push(
 )
 ```
 
+#### Existing Middleware
+
+Que ships with middleware to expose job metrics using ActiveSupport notifications to subscribe to it you can implelent the following
+
+```ruby
+::ActiveSupport::Notifications.subscribe("que_job.worked") do |message, started, finished, labels|
+  # do something with notification.
+end
+```
+
+`started` and `finished` are numeric values representing a monotonic clock so can 
+be used for timing calculations without concerning ourselves with the system clock.
+
+`labels` is a hash containing the following keys
+
+* `job_class` - the class of the job.
+* `queue` - the queue this job was queued into.
+* `priority` - the priority of this job.
+* `latency` - the amount of time this job was waiting in the queue for.
+
+To use this middleware you will have to initialize it with Que
+
+```ruby
+Que.job_middleware.push(Que::ActiveSupport::JobMiddleware)
+```
+
 ### Defining Middleware For SQL statements
 
 SQL middleware wraps queries that Que executes, or which you might decide to execute via Que.execute(). You can use hook this into NewRelic or a similar service to instrument how long SQL queries take, for example.
--- a/lib/que/active_support/job_middleware.rb
+++ b/lib/que/active_support/job_middleware.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module Que
+  module ActiveSupport
+    module JobMiddleware
+      def self.call(job)
+        labels = {
+          job_class: job.que_attrs[:job_class],
+          priority: job.que_attrs[:priority],
+          queue: job.que_attrs[:queue],
+          latency: job.que_attrs[:latency],
+        }
+
+        started = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        yield
+      ensure
+        ::ActiveSupport::Notifications.publish(
+          "que_job.worked",
+          started,
+          Process.clock_gettime(Process::CLOCK_MONOTONIC),
+          labels.merge(error: !!job.que_error.present?),
+        )
+      end
+    end
+  end
+end
--- a/lib/que/active_support/job_middleware.rb
+++ b/lib/que/active_support/job_middleware.rb
@@ -18,7 +18,7 @@ module Que
           "que_job.worked",
           started,
           Process.clock_gettime(Process::CLOCK_MONOTONIC),
-          labels.merge(error: !!job.que_error.present?),
+          labels.merge(error: job.que_error.present?),
         )
       end
     end
--- a/lib/que/poller.rb
+++ b/lib/que/poller.rb
@@ -63,6 +63,7 @@ module Que
           SELECT
             (j).*,
             l.locked,
+            extract(epoch from (now() - (j).run_at)) as latency,
             l.remaining_priorities
           FROM (
             SELECT j
@@ -81,6 +82,7 @@ module Que
             SELECT
               (j).*,
               l.locked,
+              extract(epoch from (now() - (j).run_at)) as latency,
               l.remaining_priorities
             FROM (
               SELECT
--- a/spec/que/active_support/job_middleware_spec.rb
+++ b/spec/que/active_support/job_middleware_spec.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+if defined?(::ActiveSupport)
+  require 'que/active_support/job_middleware'
+
+  describe Que::ActiveSupport::JobMiddleware do
+    let(:job) do
+      Que::Job.new(
+        job_class: "Foo",
+        priority: 100,
+        queue: "foo_queue",
+        latency: 100,
+      )
+    end
+
+    let(:labels) do
+      {
+        job_class: "Foo",
+        priority: 100,
+        queue: "foo_queue",
+        latency: 100,
+      }
+    end
+
+    it "records metrics" do
+      called = false
+      ::ActiveSupport::Notifications.subscribe("que_job.worked") do |message, started, finished, metric_labels|
+        assert_equal "que_job.worked", message
+        assert started != nil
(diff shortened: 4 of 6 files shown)


hlascelles:
Good stuff... Related: https://github.com/que-rb/que/issues/325

NB, using `now() - (j).run_at` as latency/delay can be misleading. If a job errors and retries with a backoff then the `run_at` changes. It would be best to add an "initial_run_at" column to the main jobs table, but that would be a better change....

spec/spec_helper.rb
```diff
@@ -8,6 +8,7 @@
 # in some spec runs.
 if ENV['USE_RAILS'] == 'true'
   require 'active_record'
+  require 'active_support'
```
ZimbiX, line 11: Probably a moot point if you've got Rails, but how about just `require 'active_support/notifications'`?

lib/que/active_support/job_middleware.rb
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        started = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        yield
+      ensure
+        ::ActiveSupport::Notifications.publish(
+          "que_job.worked",
+          started,
+          Process.clock_gettime(Process::CLOCK_MONOTONIC),
+          labels.merge(error: !!job.que_error.present?),
```
ZimbiX, line 21: The `!!` is redundant here, isn't it?

lib/que/active_support/job_middleware.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Que
+  module ActiveSupport
+    module JobMiddleware
```
ZimbiX, line 5: Some usage documentation would be in order, I reckon.

lib/que/poller.rb
```diff
@@ -63,6 +63,7 @@ class Poller
           SELECT
             (j).*,
             l.locked,
+            extract(epoch from (now() - (j).run_at)) as latency,
```
ZimbiX, line 66: > NB, using `now() - (j).run_at` as latency/delay can be misleading. If a job errors and retries with a backoff then the `run_at` changes. It would be best to add an "initial_run_at" column to the main jobs table, but that would be a better change....

@hlascelles This is a good point. I think the intention here is to permit ascertaining worker contention - by allowing you to analyse at the delay between when a job is desired to be run at (or re-run at), and when it's actually picked up to be run. In terms of that desire, I think this metric is more useful. But I'm sure there's also merit in https://github.com/que-rb/que/issues/325.

lib/que/active_support/job_middleware.spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    let(:labels) do
+      {
+        job_class: "Foo",
+        priority: 100,
+        queue: "foo_queue",
+        latency: 100,
+      }
+    end
```
ZimbiX, line 25: I take it you were planning to expect on these? =P

lib/que/active_support/job_middleware.rb
```diff
@@ -0,0 +1,26 @@
...
+        labels = {
+          job_class: job.que_attrs[:job_class],
+          priority: job.que_attrs[:priority],
+          queue: job.que_attrs[:queue],
+          latency: job.que_attrs[:latency],
+        }
+
+        started = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        yield
+      ensure
```
ZimbiX, line 16: I'd like a test for the exception case. But is publishing a `que_job.worked` notification even upon an exception really what you want? I'd have thought that'd mean the job is complete 🤔 There may be an example of this elsewhere..

ZimbiX (commented):
Interesting. I haven't used ActiveSupport notifications before, but knowing the job latency does sound like a useful proxy to understand the level of worker contention.

Would job latency be covered by your Prometheus metrics implementation? (Related issue: https://github.com/que-rb/que/issues/267)

I see this is still in draft. I like where this is headed. What else would be left?

Thanks for the contribution! =)

lib/que/active_support/job_middleware.rb
stephenbinns: Sounds good will add one shortly, I think so in terms of the metric were more concerned with how Que is performing rather than if the Jobs were executed correctly.

lib/que/active_support/job_middleware.rb
stephenbinns: Have added to readme do you think its better here?

lib/que/active_support/job_middleware.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module Que
+  module ActiveSupport
+    module JobMiddleware
+      def self.call(job)
+        labels = {
+          job_class: job.que_attrs[:job_class],
+          priority: job.que_attrs[:priority],
+          queue: job.que_attrs[:queue],
+          latency: job.que_attrs[:latency],
```
ameykusurkar, line 11: I think it would be also useful to have a `duration` published, so we know both how long the job was waiting in the queue, and how long it took to execute. WDYT?

stephenbinns:
> Interesting. I haven't used ActiveSupport notifications before, but knowing the job latency does sound like a useful proxy to understand the level of worker contention.
> 
> Would job latency be covered by your Prometheus metrics implementation? (Related issue: #267)
> 
> I see this is still in draft. I like where this is headed. What else would be left?
> 
> Thanks for the contribution! =)

Indeed it is part of what we're monitoring, the remaining ones are a little more tricky to figure out, as I mentioned previously we'd ideally like to get off our fork and as part of that we need to reach some sort of feature parity here.

lib/que/active_support/job_middleware.rb
stephenbinns: Execution duration can be calculated using start and end currently, we could make this more intuitive by doing that for consumers

lib/que/active_support/job_middleware.rb
ZimbiX: Ah, nah, I just wanted to start a thread so it'd be able to be marked resolved :laughing:

Cheers for adding that; looks good

docs/README.md
```diff
@@ -798,6 +798,32 @@ Que.job_middleware.push(
 )
 ```
 
+#### Existing Middleware
+
+Que ships with middleware to expose job metrics using ActiveSupport notifications to subscribe to it you can implelent the following
```
ZimbiX, line 803: *implement ;)

docs/README.md
```diff
@@ -798,6 +798,32 @@ Que.job_middleware.push(
...
+
+```ruby
+::ActiveSupport::Notifications.subscribe("que_job.worked") do |message, started, finished, labels|
```
ZimbiX, line 806: I just [looked this up](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html#module-ActiveSupport::Notifications-label-Subscribers), and it seems that `.subscribe` doesn't have monotonic timestamps; `.monotonic_subscribe` does. Whichever you want to use here is fine, but let's document it correctly.

lib/que/active_support/job_middleware.rb
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        started = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        yield
+      ensure
+        ::ActiveSupport::Notifications.publish(
+          "que_job.worked",
```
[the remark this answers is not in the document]
ZimbiX: `que.job_worked`? =P

spec/que/active_support/job_middleware_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      called = false
+      subscriber = ::ActiveSupport::Notifications.subscribe("que_job.worked") do |message, started, finished, metric_labels|
+        assert_equal "que_job.worked", message
+        assert started != nil
+        assert finished != nil
+        assert_equal labels.merge(error: true), metric_labels
+        called = true
+      end
+
+      Que::ActiveSupport::JobMiddleware.call(job) { job.que_error = "error" }
```
ZimbiX, line 47: Ah, I was thinking you'd need to raise within the block here, but you're right - an exception from a job does not bubble up to middleware.

Testing job failure is good, but I was meaning that you test that it's in an `ensure` - when it encounters an exception. If not caused by the job, perhaps by a subsequent middleware.

For realistic data, I think `job.que_error` should also be set to an instance of an exception, not that it would have any effect here.

Although.. I've just looked up the term `job_worked` in the code, and found that logging makes a distinction: `job_worked` vs `job_errored`. So maybe this middleware should do the same? Or find a more generic term - something like `job_attempted`?

ZimbiX:
Closing in favour of https://github.com/que-rb/que/pull/366

Issue #363, opened 2022-05-20
Schema version comment doesn't end up in Rails structure.sql


milgner:
While investigating possible reasons for #350 I was checking whether the DB schema was up-to-date and noticed that que [keeps it on a table comment](https://github.com/que-rb/que/blob/b674743c1599eadf088dfdaff903b1dabcc682cf/lib/que/migrations.rb#L59).

Unfortunately this comment doesn't end up in Rails' `structure.sql` so it isn't added to new systems (including `development` and `test` databases) that run `rake db:structure:load` to prepare the database.

At first glance, there seem to be two possible solutions: get Rails to dump comments (might have additional benefits because one could use them to document the schema) or introduce a separate version column or table for que (less dependencies and reliance on specific Rails behaviour).

milgner:
Looks like this was already changed in Rails:

We're currently on Rails 7.0.2 which [has `--no-comments`](https://github.com/rails/rails/blob/3520cc77df1b52a6c808083214b583c769e9a4b2/activerecord/lib/active_record/tasks/postgresql_database_tasks.rb#L61) but this line is gone in 7.0.3...  

Might be worth mentioning in the Readme, to help people running into the same issue?

milgner:
Gonna close this issue as it has been fixed in Rails 7.0.3 so it won't be a problem with new users anymore. Existing users should be able to find this issue and resolution through the search function.

ZimbiX:
I believe we just had another user get stung by this, noted in https://github.com/que-rb/que/pull/376.

Here's where the Rails bug was: [introduced](https://github.com/rails/rails/pull/43216), [resolved](https://github.com/rails/rails/pull/44633). The fix was indeed [released in Rails 7.0.3](https://github.com/rails/rails/blob/v7.0.3/activerecord/CHANGELOG.md#:~:text=remove%20--no-comments%20flag%20in%20structure%20dumps%20for%20postgresql).

Going by the Git tags shown in the [commit that introduced the issue](https://github.com/rails/rails/pull/43216/commits/d5745607463ef72adb6a34e3ecc9b3820484a33f), it looks like the issue affects every version of Rails 7 prior to 7.0.3:

![image](https://user-images.githubusercontent.com/2916331/186919810-19926c4b-a805-4d02-b3a9-145ee2c28271.png)

I wonder if we should not just document it, but add code to detect the lack of schema version comment and warn/raise!

PR #376, opened 2022-08-26
Fix DB migration name of v7 in the CHANGELOG.md

texpert:
This has been spotted when migrating to Que 2.1.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,7 @@
 This release contains a database migration. You will need to migrate Que to the latest database schema version (7). For example, on ActiveRecord and Rails 6:
 
 ```ruby
-class UpdateQueTablesToVersion6 < ActiveRecord::Migration[6.0]
+class UpdateQueTablesToVersion7 < ActiveRecord::Migration[6.0]
   def up
     Que.migrate!(version: 7)
   end


texpert:
BTW, previous, v6 DB migration, was failing on the test DB, because somehow the `pg_description` table had no row associated with the `que_jobs` relname from `pg_class`, so it was trying to run all the migration from #2.

I had to fix it with a temporary hack of setting `current = 5` here.

ZimbiX (approved):
Ah, thanks! I have to dash off, so I'll have to consider your second comment a bit later

ZimbiX:
Ah, it sounds like you might be affected by https://github.com/que-rb/que/issues/363?

texpert:
> Ah, it sounds like you might be affected by #363?

Yep, after running the migration, the comment with the version has appeared - https://github.com/texpert/rails_6_rss_reader/commit/383af1e662d31fd63d895c4713fde478a9fab76c#diff-1dd5a8f580b9615769ddc19a64a98f455d9d20eb149c443f41d9d3c797d39b12R68

Still, this could be a problem for the old users migrating.

ZimbiX:
Damn, ok. Yeah, I'm [thinking we should add something to detect this issue](https://github.com/que-rb/que/issues/363#issuecomment-1228525503).

Issue #377, opened 2022-08-26
Detect Que schema version missing due to bug in Rails schema dump - affects all versions of Rails 7 prior to 7.0.3


ZimbiX:
See https://github.com/que-rb/que/issues/363#issuecomment-1228525503.

I reckon we should add some code to detect this at startup and warn/raise.

It could be possible to automatically detect which version of the schema we're migrated to in order to populate the missing comment automatically, but we'd have to maintain that with each new migration version. I don't think that's sustainable, and would defeat the purpose of recording the schema version anyway.

This could probably be addressed at the same time as https://github.com/que-rb/que/issues/375.

PR #379, opened 2022-08-26
Raise when DB schema version comment is missing

ZimbiX:
Resolves https://github.com/que-rb/que/issues/377.

[PostgreSQL comment docs](https://www.postgresql.org/docs/current/sql-comment.html)

~~This wasn't as simple as I first thought it'd be, since we also have to cater for the scenario of migrating from before DB schema version was recorded. To detect that case, I'm checking the default for the `priority` column, since that was changed to a much more sensible value in [migration 2](https://github.com/que-rb/que/blob/0cdd37c082443ceef00bdbb10ed5f74b7fe6efbb/lib/que/migrations/2/up.sql).~~

~~I did some code archaeology, and determined that, prior to migration 2, the default for the `priority` column was 1 since https://github.com/que-rb/que/commit/1050850204629958a7c5a7bd5ae0076ce56dd6a2#diff-f5ed5e5d3159a4a7d9d7ea114936cb2147d07e27ce81dcce79c2814e299d2456R8, which (according to the tags that commit appears in) covers way back to v0.1.0. So this is a safe proxy to detect DB schemas pre version 2. However, the `que_jobs` table schema at the linked commit differs from the current [migration 1](https://github.com/que-rb/que/blob/0cdd37c082443ceef00bdbb10ed5f74b7fe6efbb/lib/que/migrations/1/up.sql) - to consider it as schema version 1 and migrate up would produce an incorrect result, if it even works. But regardless, this check against the `priority` column's default preserves the DB schema version 1 detection result as it was before.~~

Migrating when the DB schema version comment is missing now looks like:

```
➜ docker compose up -d db

➜ export DATABASE_URL=postgres://que:que@localhost:5697/que-test

➜ ruby -r bundler/setup -r que -r sequel -e "Que.connection = Sequel.connect(ENV['DATABASE_URL']); Que.migrate!(version: Que::Migrations::CURRENT_VERSION)"

➜ ./auto/psql -c "comment on table que_jobs is null;" 
[+] Running 1/0
 ⠿ Container que-db-1  Running  0.0s
COMMENT

➜ ruby -r bundler/setup -r que -r sequel -e "Que.connection = Sequel.connect(ENV['DATABASE_URL']); Que.migrate!(version: Que::Migrations::CURRENT_VERSION)"
/home/brendan/Projects/que/lib/que/migrations.rb:78:in `_raise_db_version_comment_missing_error': Cannot determine Que DB schema version. (Que::Error)

The que_jobs table is abnormally missing its comment recording the Que DB schema version. This is likely due to a bug in Rails schema dump in Rails 7 versions prior to 7.0.3, omitting comments - see https://github.com/que-rb/que/issues/377. Please determine the appropriate schema version from your migrations and record it manually by running the following SQL (replacing version as appropriate):

COMMENT ON TABLE que_jobs IS 'version';
  from /home/brendan/Projects/que/lib/que/migrations.rb:55:in `db_version'
  from /home/brendan/Projects/que/lib/que/migrations.rb:12:in `block in migrate!'
  from /home/brendan/Projects/que/lib/que/utils/transactions.rb:18:in `block in transaction'
  from /home/brendan/Projects/que/lib/que/connection_pool.rb:39:in `block in checkout'
  from /home/brendan/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/sequel-5.58.0/lib/sequel/connection_pool/threaded.rb:92:in `hold'
  from /home/brendan/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/sequel-5.58.0/lib/sequel/database/connecting.rb:269:in `synchronize'
  from /home/brendan/Projects/que/lib/que/connection_pool.rb:18:in `checkout'
  from /home/brendan/Projects/que/lib/que/utils/transactions.rb:12:in `transaction'
  from /home/brendan/Projects/que/lib/que/migrations.rb:11:in `migrate!'
  from /home/brendan/.rbenv/versions/3.1.2/lib/ruby/3.1.0/forwardable.rb:238:in `migrate!'
  from -e:1:in `<main>'

➜ docker compose down && docker volume rm -f que_db-data
```

### DIFF
--- a/lib/que/migrations.rb
+++ b/lib/que/migrations.rb
@@ -77,7 +77,7 @@ module Que
         raise Error, <<~ERROR
           Cannot determine Que DB schema version.
 
-          The que_jobs table is abnormally missing its comment recording the Que DB schema version. This is likely due to a bug in Rails schema dump in Rails 7 versions prior to 7.0.3, omitting comments - see https://github.com/que-rb/que/issues/377. Please determine the appropriate schema version from your migrations and record it manually by running the following SQL (replacing version as appropriate):
+          The que_jobs table is abnormally missing its comment recording the Que DB schema version. This is likely due to a bug in Rails schema dump in Rails 7 versions prior to 7.0.3, omitting comments - see https://github.com/que-rb/que/issues/363. Please determine the appropriate schema version from your migrations and record it manually by running the following SQL (replacing version as appropriate):
 
           COMMENT ON TABLE que_jobs IS 'version';
         ERROR
--- a/lib/que/migrations.rb
+++ b/lib/que/migrations.rb
@@ -48,12 +48,39 @@ module Que
           # No table in the database at all.
           0
         elsif (d = result.first[:description]).nil?
-          # There's a table, it was just created before the migration system
-          # existed.
-          1
+          # The table exists but the version comment is missing
+          if _db_schema_matches_db_version_1?
+            1
+          else
+            _raise_db_version_comment_missing_error
+          end
         else
           d.to_i
         end
+      end
+
+      # The que_jobs table could be missing the schema version comment either due to:
+      # - Being created before the migration system existed (matching DB version 1); or
+      # - A bug in Rails schema dump in some versions of Rails
+      # So to determine which is the case, here we check an aspect of the schema that's only like that if just the first migration has been applied
+      def _db_schema_matches_db_version_1?
+        result =
+          Que.execute <<-SQL
+            SELECT column_default AS default_priority
+            FROM information_schema.columns
+            WHERE (table_schema, table_name, column_name) = ('public', 'que_jobs', 'priority');
+          SQL
+        result.first[:default_priority] == '1'
+      end
+
+      def _raise_db_version_comment_missing_error
+        raise Error, <<~ERROR
+          Cannot determine Que DB schema version.
+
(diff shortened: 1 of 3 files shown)


lib/que/migrations.rb
```diff
@@ -48,14 +48,41 @@ def db_version
...
         elsif (d = result.first[:description]).nil?
-          # There's a table, it was just created before the migration system
-          # existed.
-          1
+          # The table exists but the version comment is missing
+          if _db_schema_matches_db_version_1?
+            1
+          else
+            _raise_db_version_comment_missing_error
+          end
```
oeoeaio, line 56: Is there really a use case for continuing to support the version 1 schema? It’s going to be fundamentally incompatible with the rest of the code base at this point. Surely we can just raise if the comment is missing?

ZimbiX: Yeah, I guess not, actually. The migrations system was introduced in [v0.5.0](https://github.com/que-rb/que/blob/master/CHANGELOG.md#050-2014-01-14) which came out on 2014-01-14. With the latest version of 0.x being v0.14.3 (2018-03-02), I guess we can assume no one's on a version below v0.5.0.

I'll change the PR to assume DB schema version 1 would have the comment, and just add a note in the changelog that support for upgrading directly from versions below v0.5.0 is unsupported.

maddymarkovitz (approved):

PR #380, opened 2022-08-31
Make the current Locker instance available via Que.locker

mnbbrown:
This makes the current running `Locker` instance accessible externally. This allows other code to inspect the state of the running locker, its workers and associated threads.

See https://github.com/que-rb/que/pull/366#issuecomment-1229466410 for more context.

### DIFF
--- a/lib/que/locker.rb
+++ b/lib/que/locker.rb
@@ -7,6 +7,10 @@
 require 'set'
 
 module Que
+  class << self
+    attr_accessor :locker
+  end
+
   Listener::MESSAGE_FORMATS[:job_available] =
     {
       queue:    String,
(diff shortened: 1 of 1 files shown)


lib/que/locker.rb
```diff
@@ -71,6 +75,11 @@ def initialize(
       Que.assert Array, worker_priorities
       worker_priorities.each { |p| Que.assert([Integer, NilClass], p) }
 
+      # We assign this globally because we only ever expect on locker to be
```
ZimbiX, line 78: *one locker

ZimbiX (changes requested):
Looks good otherwise! =)

lib/que/locker.rb
mnbbrown: How did I miss that :O fixed now :)

ZimbiX (approved):

PR #382, opened 2022-09-08
Allow library users to check if Que is running via CLI or lib

mnbbrown:
This follows the sidekiq pattern (i.e. `Sidekiq.server? == true` when running as a worker)
Useful for code to understand if it's running in the `bundle exec que` worker context or not.

Why? 

We'd like to be able to determine within the rails code base (i.e. `config/environment.rb` onwards) if the code is running in the context of the the `que` CLI, or if it's running in the normal rails context. For example, we have some Rails initializers that should run under normal circumstances but not run when running in a Que worker:

```ruby
# config/initializers/custom.rb

if !Que.server?
   # setup up some things
end
```

### DIFF
--- a/lib/que.rb
+++ b/lib/que.rb
@@ -79,6 +79,10 @@ module Que
       @default_queue || DEFAULT_QUEUE
     end
 
+    def server?
+      defined?(Que::CommandLineInterface).nil?
+    end
+
     # Support simple integration with many common connection pools.
     def connection=(conn)
       self.connection_proc =


ZimbiX:
Hmm, I'm not sure I understand how this is useful - what are you planning to use it for?

mnbbrown:
> Hmm, I'm not sure I understand how this is useful - what are you planning to use it for?

Yeah - the description is poor - apologies. I've updated it with a rationale.

mnbbrown:
> Hmm, I'm not sure I understand how this is useful - what are you planning to use it for?

Hey @ZimbiX - just checking in if the description is more useful now?

lib/que.rb
```diff
@@ -79,6 +79,10 @@ def default_queue
       @default_queue || DEFAULT_QUEUE
     end
 
+    def server?
+      defined?(Que::CommandLineInterface).nil?
```
ZimbiX, line 83: I haven't seen `defined?` before - I've always used `Kernel.const_defined?("MyClass")`. Handy!

ZimbiX (approved):
Thanks for the extra info. That makes sense now.

If this is supported, I think it would be appropriate to document how to use Que that way - instantiating a locker inside another Ruby process, e.g. Rails / puma. I'll merge this, but not release just yet, I think; can you contribute quick docs for what you know in another PR? Then I can release that together.

PR #389, opened 2023-03-31
fix(SynchronousJobs) Fix synchronous jobs return value

Lomig:
# Context

* We use `Que` in my company as one of a few queue providers to manage jobs in different services through a custom framework abstracting its usage
* For testing purposes, we use `Que` in synchronous mode
* We stumbled upon an issue where failed synchronous jobs were raising unexpected errors when trying to gather metrics and logs about their status, whereas it was working perfectly fine in normal production (asynchronous) state

Screenshot 2023-03-31 at 14 00 06

The PR aims at fixing this issue without the boilerplate in our codebase to mitigate this issue which exists just during testing.

# Description
   * Asynchronous jobs return themselves once ran
   * Synchronous jobs return the result of their call

This PR makes both calls returning the same thing, which is the job itself.

### DIFF
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -246,6 +246,7 @@ module Que
         new(attrs).tap do |job|
           Que.run_job_middleware(job) do
             job._run(reraise_errors: true)
+            job
           end
         end
       end


oeoeaio:
Hi @Lomig, thanks for your PR. I'd like to see some tests covering the behaviour that you're expecting.

To be honest I'm not sure I understand how you're using the return values of these methods, so a little more detail about that might be me understand your use case. Maybe just adding the tests will be enough.

Lomig:
Hi!

I would be glad to propose some tests once we agree of the need – for now there is no test around what I propose indeed!


Here is the use case:
As I described, we use `que` in an custom framework, and use several pieces of job middleware to add custom instrumentation.

Particularly here, we want to gather metrics on successful jobs and catch errors on any issue that could be raised. the middleware is as follows (simplified):

```ruby
Que.job_middleware.push(lambda do |job, &block|
  result = nil

  duration = Benchmark.realtime do
    result = block.call
  end
  @prometheus.observe(duration)
          
  if result&.que_resolved && !result.que_error
    @logger.info_data(
      {
        job_id: job.que_attrs[:id],
        job_type: job.class.to_s,
        duration: duration,
        error_count: job.que_attrs[:error_count],
        args: job.que_attrs[:args].to_json,
        kwargs: job.que_attrs[:kwargs].to_json
      },
      "Job successful"
    )

    @prometheus.increment(:job_successful)
  end

  rescue Exception => e
    handle_error(e, job.que_attrs)
end)
```

This part `if result&.que_resolved && !result.que_error` only works for synchronous calls because of what I described in the PR description.

Maybe it could have been implemented in another fashion, as I'm just discovering this part of the code base, and also as I am the first to have decided for synchronous jobs during testing, and thus I am not that familiar with Que 😅

In any case, thanks for your answer ☺️

oeoeaio:
Thanks @Lomig I think I understand a little more now.

Maybe I'm missing something, but `que_resolved` and `que_error` are both available on the `job` itself.

Does this work for you?

```ruby
Que.job_middleware.push(
  lambda do |job, &block|
    duration = Benchmark.realtime do
      block.call
    end
    @prometheus.observe(duration)
          
    if job.que_resolved && !job.que_error
      @logger.info_and_metadata(
        "Job successful",
        {
          job_id: job.que_attrs[:id],
          job_type: job.class.to_s,
          duration: duration,
          error_count: job.que_attrs[:error_count],
          args: job.que_attrs[:args].to_json,
          kwargs: job.que_attrs[:kwargs].to_json
        }
      )

      @prometheus.increment(:job_successful)
    end
  rescue Exception => e
    handle_error(e, job)
  end
)
```

Lomig:
I cannot reach the dev who created this piece of code, so I don't know why he used the returned value of the block instead of the job itself.

I didn't challenge this decision from before my time (and I should have, as you demonstrated), and assumed that the block call was returning a new job object and not mutating the initial one.
If there is mutation indeed, using the job directly is a way better solution indeed!

I'll close this PR then. Sorry for having bothered you 😖

PR #390, opened 2023-04-12
Fix support for ActiveJob in Ruby 3.2

kevinrobayna:
Resolves https://github.com/que-rb/que/issues/391

Ruby 2.7 has reached end of live so this leaves the supported versions being 3.0, 3.1 and the latest one 3.2 [source](https://www.ruby-lang.org/en/news/2023/03/30/ruby-2-7-8-released/)


ZimbiX (commented):
Cheers. Looks good so far.

Although Ruby 2.7 has been end of life'd, I don't think we should drop it yet. The upgrade process to Ruby 3 is often not trivial, and not everyone has upgraded. And I believe in being a bit more forgiving with older versions, in my career having worked on several legacy apps lacking upgrade budget.

kevinrobayna:
> Cheers. Looks good so far.
> 
> Although Ruby 2.7 has been end of life'd, I don't think we should drop it yet. The upgrade process to Ruby 3 is often not trivial, and not everyone has upgraded. And I believe in being a bit more forgiving with older versions, in my career having worked on several legacy apps lacking upgrade budget.

yeah totally agree! I think the path forward should be to ensure we run with the latest but keep running with older versions and only break backward compatibility on purpose and not by accident

oeoeaio:
Thanks for your PR @kevinrobayna. Just so I am clear, are you intending to make changes to the implementation too?

ZimbiX:
@oeoeaio See [on Discord](https://discord.com/channels/625638970327433217/628151201598537740/1095767657921400852):

> I had no idea there'd be a problem with Ruby 3.2 - thanks for bringing this to our attention!
> I think I've worked out the solution 🙂
> https://github.com/que-rb/que/commit/4c8937ccbeadda16b1a9d0ae528bf62e835f2fc7
> I think this Ruby 3.2 & ActiveJob problem is worthy of having a GitHub issue filed (on the Que repo). Can you do that? Then you can incorporate my commit into your PR and mark the PR as resolving the issue

ZimbiX (changes requested):
Given we're adding Ruby 3.2 here, and changing the default in `.ruby-version`, I think it's an appropriate time to update everything to use 3.2:

- The Ruby version used in the [GitHub Actions tests of each Postgres version](https://github.com/que-rb/que/blob/d4c0cae39a6c3101af16103bfbcbbab8c55a9cef/.github/workflows/tests.yml#L15-L20)
- The Ruby base Docker image in the [`Dockerfile`](https://github.com/que-rb/que/blob/a82e12eaec8ccf674c16cc9e3344e56900fc83c2/Dockerfile#L1)
- The gem cache volume name in the [`docker-compose.yml`](https://github.com/que-rb/que/blob/a82e12eaec8ccf674c16cc9e3344e56900fc83c2/docker-compose.yml#L13)

ZimbiX (approved):
@oeoeaio Does my fix look alright to you? 1d1201db176b8ce6244e3f3ba9f8c24754ffce25. It passes on all versions on GitHub Actions at least

spec/que/job_spec.rb
```diff
@@ -83,11 +89,13 @@ def expected_job_count
           enqueue_method.call({'a' => 1, 'b' => 2}, c: 3, d: 4)
           execute
           assert_equal([{a: 1, b: 2}], $args)
+          assert_equal({:c=>3, :d=>4}, $kwargs)
```
oeoeaio, line 92: Any reason to use hash rockets rather than colons here?

ZimbiX: Ah, we don't have RuboCop yet. Yeah, let's use colon syntax for consistency

lib/que/active_job/extensions.rb
```diff
@@ -31,6 +32,7 @@ def perform(*args)
           ),
         )
       end
+      ruby2_keywords(:perform) if respond_to?(:ruby2_keywords, true)
```
oeoeaio, line 35: > Does my fix look alright to you?

Yeah @ZimbiX, makes sense to me! :)

spec/que/job_spec.rb
```diff
@@ -33,11 +33,13 @@ def expected_job_count
           enqueue_method.call(1, 2)
           execute
           assert_equal([1, 2], $args)
+          assert_equal({}, $kwargs)
         end
 
         it "should pass its keyword arguments to the run method" do
           enqueue_method.call(a: 1, b: 2)
           execute
+          assert_equal([], $args)
```
oeoeaio, line 42: Just checking, do we need all of the assertions below this line?

spec/que/job_spec.rb
```diff
@@ -83,11 +89,13 @@ def expected_job_count
...
           execute
           assert_equal([{a: 1, b: 2}], $args)
+          assert_equal({:c=>3, :d=>4}, $kwargs)
         end
 
         it "should symbolize keyword argument keys" do
           enqueue_method.call(a: 1, b: 2)
           execute
+          assert_equal([], $args)
           assert_equal({a: 1, b: 2}, $kwargs)
```
oeoeaio, line 99: This new assertion doesn't really match the example description (which talks about keyword args). Is this assertion strictly necessary (not redundant) and if so can it be moved into a separate example?

ZimbiX: With how the args/kwargs separation was messing up, I can see why you'd want to assert on both. We do that in the bulk enqueue spec, just through a helper:

https://github.com/que-rb/que/blob/a82e12eaec8ccf674c16cc9e3344e56900fc83c2/spec/que/job.bulk_enqueue_spec.rb#L77-L78

https://github.com/que-rb/que/blob/a82e12eaec8ccf674c16cc9e3344e56900fc83c2/spec/que/job.bulk_enqueue_spec.rb#L30-L31

oeoeaio: Yeah, except those examples aren't quite as specific. I'm happy to waive this through but keen to understand whether there is genuinely no redundancy in the tests here or if we're just fleshing them out to consistently make assertions about both `args` and `kwargs`.

ZimbiX: I can see where you're coming from too. I don't know :shrug: I'm happy either way

kevinrobayna: Yeah I think personally this assertion is just to provide documentation to the reader. It's def not needed but I feel it's nice.

kevinrobayna:
@ZimbiX I feel this PR is soon ready to be merged unless we want to remove the double assertion of what happens on args and kwargs

ZimbiX:
Let's give @oeoeaio final say here

oeoeaio (approved):

PR #392, opened 2023-04-12
Add support for rails version 7.0

kevinrobayna:
Rails 7.0 was released a while ago so we should ensure we support it.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,16 +8,16 @@ jobs:
     strategy:
       matrix:
         ruby_version: ['2.7', '3.0', '3.1', '3.2']
-        rails_gemfile: ['6.0', '6.1']
+        rails_gemfile: ['6.0', '6.1', '7.0']
         postgres_version: ['14']
         include:
         # Postgres versions
-        - { ruby_version: '3.2', rails_gemfile: '6.1', postgres_version: '9' }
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         ruby_version: ['2.7', '3.0', '3.1']
-        rails_gemfile: ['6.0', '6.1']
+        rails_gemfile: ['6.0', '6.1', '7.0']
```
kevinrobayna, line 11: 6.0 is reaching its end of life soon but again if it's not annoying anyone we should keep it. https://endoflife.date/rails

ZimbiX (changes requested):
Cheers =)

This will need to be rebased once https://github.com/que-rb/que/pull/390 merges.

Again, let's update the Postgres versions tests to use the latest of everything else.

We're getting a lot of matrix permutations again (testing a version of something against every version of everything else). We don't have to address this now, but if a build starts to take too long to get through them all, we should probably move some out like with the PostgreSQL versions.

oeoeaio:
I will hold off doing a patch release until this is rebased and merged.

kevinrobayna:
Hi @ZimbiX and @oeoeaio i just rebased the branch let see what the specs say.

ZimbiX (approved):
Cheers

PR #395, opened 2023-05-19
Allow que to be started without listen/notify

dtcristo:
Ability to disable listen/notify on CLI resulting in jobs being picked up purely from polling. We have seen issues with listen/notify usage at scale.

### DIFF
--- a/lib/que/command_line_interface.rb
+++ b/lib/que/command_line_interface.rb
@@ -51,6 +51,14 @@ module Que
             end
 
             opts.on(
+              '--listen [LISTEN]',
+              String,
+              "Set to false to disable listen mode (default: true)"
+            ) do |listen|
+              options[:listen] = listen != "false"
+            end
+
+            opts.on(
               '-l',
               '--log-level [LEVEL]',
               String,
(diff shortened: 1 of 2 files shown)


oeoeaio:
Thanks for this! No idea why this is not already configurable. If you can add some specs I'll be happy to approve and merge this. :)

dtcristo:
@oeoeaio I've updated the commit with a spec for the new option. Ready for review.

mnbbrown:
Do you think we could get this merged?
@dtcristo / @ZimbiX shout if I can help in anyway.

oeoeaio (approved):

ZimbiX:
Sorry for the delay!

Out of curiosity, what issues were you seeing with it at scale?

dtcristo:
@oeoeaio Thanks for merging!

> Out of curiosity, what issues were you seeing with listen/notify at scale?

Hey @ZimbiX, we have quite a high throughput of jobs in our system. Average of 150 jobs/s peaking to about 700 jobs/s at times. We had issues where listen/notify queue would fill quickly and when full subsequent listen/notify calls would fail. I believe there were issues where long running transactions prevented the queue from being emptied if the transactions covered items in the queue.

I feel listen/notify is ideal for systems with lower job throughput with sporadic events and you need low latency. In a system with high throughput like ours, poll mode works fine.

PR #396, opened 2023-05-19
Prevent jobs from being enqueued in the past

dtcristo:
It's relatively easy for this to happen by accident (e.g. time drift on a rogue server), and it throws a spanner in the works if you're monitoring for oldest job in the queue.

### TODO
- Fix specs.

### DIFF
--- a/lib/que/job.rb
+++ b/lib/que/job.rb
@@ -17,7 +17,7 @@ module Que
         (
           coalesce($1, 'default')::text,
           coalesce($2, 100)::smallint,
-          coalesce($3, now())::timestamptz,
+          greatest($3, now())::timestamptz,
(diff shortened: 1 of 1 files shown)


hlascelles:
We wouldn't like this. If we have many similar jobs (same queue and priority) we can use run_at to put a job to the front. 

Also if you have (say) a schedule you'd like to insert at the beginning of the day, distributed across the day, then if they back up they will run in a predictable order. If this change goes in and the scheduler runs late, then the run order is non-deterministic.

oeoeaio:
I have to say that I'm not in favour of enforcing this kind of behaviour universally. If you need this functionality I suggest you build it into some kind of wrapper.

PR #401, opened 2023-10-09
Allow running Que on Rails 7.1

texpert:
QueAdapter [has been removed from ActiveJob in Rails 7.1](https://github.com/rails/rails/pull/46005), so now a Rails 7.1 app fails to start with the `cannot load such file -- active_job/queue_adapters/que_adapter (LoadError)` even if not using ActiveJob with Que.

This PR is enabling running Que on Rails 7.1 not requiring `que/active_job/extensions`.

Issue #399, opened 2023-09-19
Is there any way to use tags with ActiveJob


skv-headless:
Tags work:
`ChargeCreditCard.enqueue(card.id, user_id: current_user.id, job_options: { tags: ["hello"] })`

Tags don't work:
`ChargeCreditCard.perform_later(card.id, user_id: current_user.id, job_options: { tags: ["hello"] })`

Maybe I miss something?

ZimbiX:
Hmm, I'm afraid not!

https://github.com/que-rb/que/blob/9744019261eccc4bd3a2f3d05d45c957632c2d22/lib/que/active_job/extensions.rb#L62-L70

I'll have a quick look into how tags support could be added.

ZimbiX:
I've not used tags before, nor have I ever been clear on their purpose. Is there a reason you can't have your tags be a job kwarg?

ZimbiX:
> We've dropped support for job options supplied as top-level keywords, but ActiveJob's QueAdapter still uses them. So we have to move them into the job_options hash ourselves. 

I looked into this. The manual construction of `job_options` was only in edge Rails for 6 months (https://github.com/rails/rails/pull/44734 until https://github.com/rails/rails/pull/46005 / 2022-03-22 – 2022-09-13), with both merge commits showing a Git tag of `v7.1.0.beta1`, so given that, I read that it never made it into the `v7.1.0` release. However, it [was backported to `7-0-stable`](https://github.com/rails/rails/pull/44734#issuecomment-1159380677), and it's [still present there](https://github.com/rails/rails/blob/7-0-stable/activejob/lib/active_job/queue_adapters/que_adapter.rb), so I gathered that it was released in a v7.0.x, and yes, it was [released in v7.0.4](https://github.com/rails/rails/releases/tag/v7.0.4#:~:text=Jakub%20Malinowski-,active%20job,-Update%20ActiveJob%3A%3AQueueAdapters).

So. We can't just assume it's safe to change the handling of `job_options` in Que's ActiveJob extensions.

skv-headless:
> I've not used tags before, nor have I ever been clear on their purpose.
for example synchronization produces some amount of jobs. I mark them with tag and get synchronization progress.

> Is there a reason you can't have your tags be a job kwarg
I think I can. It would just require some refactoring

Wanted your opinion before starting to refactor. 
Thank you for a quick response 🙏

ZimbiX:
[...]

PR #403, opened 2023-10-09
Support Rails 7.1+ by adding ActiveJob Que adapter removed from Rails

ZimbiX:
ActiveJob's Que adapter was removed from Rails in https://github.com/rails/rails/pull/46005 (issue: https://github.com/rails/rails/issues/45899). This is a good thing for simplicity - Que and ActiveJob won't have to keep patching over each other's changes. But it's unfortunately broken Que in Rails.

Sorry, I didn't realise this would be an issue in removing the Que adapter from Rails! 😬 I really should have tested that before giving the go ahead.

The adapter is still necessary, and so this PR incorporates it into Que.

I'd thought we'd need to now use `ActiveJob::Base.queue_adapter = ActiveJob::QueueAdapters::QueAdapter.new` rather than `ActiveJob::Base.queue_adapter = :que`, but it seems that by keeping the adapter class namespace unchanged, we can still use the latter. We have to keep it unchanged for backwards compatibility, as the class name of `ActiveJob::QueueAdapters::QueAdapter::JobWrapper` must remain the same, given this string would be in old job records in the database.

Because the class name needs to remain the same, but we still need to support older versions of ActiveJob where that class exists there, I've defined the adapter here only when the version of ActiveJob is 7.1+.

I was able to simplify the adapter slightly by removing `#require_job_options_kwarg?`, since the Que codebase only has to cater for one version of Que =P

I considered removing & inlining `Que::ActiveJob::WrapperExtensions::ClassMethods#enqueue` into our Rails 7.1+ QueAdapter, but it still needs to exist in place to support older versions of Rails =\ Keeping it in place conditionally based on the Rails version would be [a bit complex](https://github.com/que-rb/que/issues/399#issuecomment-1725170442), and so perhaps not an improvement.

Supersedes https://github.com/que-rb/que/pull/401.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,16 +8,16 @@ jobs:
     strategy:
       matrix:
         ruby_version: ['2.7', '3.0', '3.1', '3.2']
-        rails_gemfile: ['6.0', '6.1', '7.0']
+        rails_gemfile: ['6.0', '6.1', '7.0', '7.1']
         postgres_version: ['14']
         include:
         # Postgres versions
-        - { ruby_version: '3.2', rails_gemfile: '7.0', postgres_version: '9' }
-        - { ruby_version: '3.2', rails_gemfile: '7.0', postgres_version: '10' }
-        - { ruby_version: '3.2', rails_gemfile: '7.0', postgres_version: '11' }
-        - { ruby_version: '3.2', rails_gemfile: '7.0', postgres_version: '12' }
-        - { ruby_version: '3.2', rails_gemfile: '7.0', postgres_version: '13' }
-        - { ruby_version: '3.2', rails_gemfile: '7.0', postgres_version: '14' }
+        - { ruby_version: '3.2', rails_gemfile: '7.1', postgres_version: '9' }
+        - { ruby_version: '3.2', rails_gemfile: '7.1', postgres_version: '10' }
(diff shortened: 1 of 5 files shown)


texpert (approved):
Awesome!

mezza:
Any ETA on the PR being merged and a new gem being cut? Thanks in advance.

oeoeaio (approved):

oeoeaio:
Thanks @ZimbiX, let's do it.

oeoeaio:
Gah, unfortunately this has broken the build. Trying to work out why. Maybe Rails 7.1.1 release a few days ago breaks support for postgres 9 somehow?

oeoeaio:
> Any ETA on the PR being merged and a new gem being cut? Thanks in advance.

Done. @ZimbiX I might need to you announce it on discord, I don't seem to have permission.

ZimbiX:
> @ZimbiX I might need to you announce it on discord, I don't seem to have permission.

Done =)

Thanks for working out the failure! I agree with your workaround (https://github.com/que-rb/que/pull/405). I imagine their fix would be included in the next release of Rails, so it looks like [excluding 7.1.1](https://www.simplybusiness.co.uk/about-us/tech/2017/11/bundler-can-exclude-gem-versions/) was another option.

PR #404, opened 2023-10-12
Allow que to be started and create pid file

jokius:
It's help when you need make heals check in kubernetes

### DIFF
--- a/lib/que/locker.rb
+++ b/lib/que/locker.rb
@@ -61,7 +61,8 @@ module Que
       wait_period:         DEFAULT_WAIT_PERIOD,
       maximum_buffer_size: DEFAULT_MAXIMUM_BUFFER_SIZE,
       worker_priorities:   DEFAULT_WORKER_PRIORITIES,
-      on_worker_start:     nil
+      on_worker_start:     nil,
+      pidfile:             nil
     )
 
       # Sanity-check all our arguments, since some users may instantiate Locker
@@ -215,10 +216,16 @@ module Que
             @connection.wrapped_connection.close
           end
         end
+
+      @pidfile = pidfile
+      at_exit { delete_pid }
+      write_pid
     end
 
     def stop!
-      stop; wait_for_stop
+      stop
+      wait_for_stop
+      delete_pid
     end
 
     def stop
@@ -511,6 +518,20 @@ module Que
       Que.assert(@locks.add?(id)) do
         "Tried to lock a job that was already locked: #{id}"
       end
+    end
+
+    def write_pid
+      return unless @pidfile
+
+      File.open(@pidfile, "w+") do |f|
+        f.write(Process.pid.to_s)
+      end
+    end
+
+    def delete_pid
+      return unless @pidfile
+
+      File.delete(@pidfile) if File.exist?(@pidfile)
     end
   end
 end
--- a/spec/que/locker_spec.rb
+++ b/spec/que/locker_spec.rb
@@ -185,6 +185,30 @@ describe Que::Locker do
     locker.stop!
   end
 
+  describe "with pidfile" do
+    it "should create and delete pid file, if not exist" do
+      pidfile = "./spec/temp/pidfile_#{Time.new.to_i}.pid"
+      locker_settings[:pidfile] = pidfile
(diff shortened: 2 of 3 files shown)


jokius:
@oeoeaio Now yes.

spec/que/locker_spec.rb
```diff
@@ -185,6 +185,32 @@ def assert_que_locker_insertion(
...
+    File.stub :open, mock_open do
+      File.stub :write, mock_file do
+        locker
+        locker.stop!
+      end
+    end
+
+    File.stub :delete, mock_delete do
+      system("ruby -e 'require \"minitest/autorun\"; exit Minitest.run(ARGV)'")
+    end
```
ZimbiX: Testing `at_exit` doesn't seem straightforward.

I think I see what you're trying to do here - have a process run the `at_exit` block to delete the file. But doesn't this `system` call run all the tests again? If that even works, I don't think this is a good way to trigger the file deletion - it's indirect and very slow.

How about extracting `#write_pid` into a new class, and testing that in isolation in a similar way to what you were trying to do by using a subprocess, e.g.:

```ruby
# Start a bakground process in a thread:
thread = Thread.new do
  `ruby -r ./lib/que/pid_file.rb -e 'Que::PidFile.create('testfile'); sleep'`
end
# Then check pidfile is present (will require polling and a timeout, since it's beign created asynchronously)
# Then kill the thread, which should close the subprocess
# Then check pidfile is not present
```

jokius: @ZimbiX No, that doesn't work. When you kill the thread, ruby does not call at_exit. But I changed the thread and now everything works fine.

@oeoeaio Now all tests are green.

jokius:
@oeoeaio Now all tests are green.

lib/que/locker.rb
```diff
@@ -215,10 +216,16 @@ def initialize(
             @connection.wrapped_connection.close
           end
         end
+
+      @pidfile = pidfile
+      write_pid
+      at_exit { delete_pid } # Guaranty delete the pidfile on exit or crashed.
```
ZimbiX, line 222: I think setting up this deletion before writing the file would be ever so slightly more robust - move this line up one, I reckon. And I don't think the comment is necessary - it's pretty clear what this is for.

spec/que/locker_spec.rb
ZimbiX: I'm sure there would have been a way to get something like that to work - like manually sending a particular kill signal to the subprocess's pid. But I think your alternative is fine

lib/que/locker.rb
```diff
@@ -512,5 +519,19 @@ def mark_id_as_locked(id)
         "Tried to lock a job that was already locked: #{id}"
       end
     end
+
+    def write_pid
+      return unless @pidfile
+
+      File.open(@pidfile, ::File::CREAT | ::File::EXCL | ::File::WRONLY) do |f|
```
ZimbiX, line 526: I haven't seen these options given to file opening before - what are they doing? / why are they necessary here? To cater for if the file already exists? If it's important, maybe it's worth adding a test?

spec/que/locker_spec.rb
```diff
@@ -185,6 +185,16 @@ def assert_que_locker_insertion(
     locker.stop!
   end
 
+  it "should create and delete pid file, if passed" do
+    pidfile = "./spec/temp/pidfile_#{Time.new.to_i}.pid"
+    locker_settings[:pidfile] = pidfile
+    locker
+    assert_equal true, File.exist?(pidfile)
```
ZimbiX, line 192: Style suggestion: you can instead call `assert` and `refute` for these two boolean assertions

ZimbiX (changes requested):
Thanks for the improvements =)

lib/que/locker.rb
jokius: You right. Change a little :)

ZimbiX (approved):
Looks good. Cheers!

dkoprov:
Hey! Any plans to release this patch into a new release? 2.4.0 maybe? Thank you.

PR #406, opened 2023-10-19
Add view and job table enhancements for improved observability

maxschridde1494:
Changes include:

**Additional que_jobs.first_run_at column**
The `first_run_at` column will store the timestamp with time zone (timestamptz) of the initial scheduled execution time for jobs. The column default is set to now, but realistically this will always default to the initial `run_at` value. This enhancement helps re-trace the execution of jobs when reviewing failures.

**NEW que_jobs_ext view**
This view extends the functionality of the job management system by providing an enriched view of jobs. It combines data from the `que_jobs` table and the `pg_locks` table to present a comprehensive overview of jobs, including their status and locking details. Here is an example SQL command highlighting usage of the new view:

```SQL
SELECT
  count(*) FILTER(where status='scheduled') as scheduled
  , count(*) FILTER(where status='queued') as queued
  , count(*) FILTER(where status='running') as running
  , count(*) FILTER(where status='failed') as failed
  , count(*) FILTER(where status='errored') as errored
  , count(*) FILTER(where status='completed') as completed
FROM que_jobs_ext
```

We have used these new enhancements to power some of our admin dashboards. Below is a video of one such example.

[IMAGE ALT TEXT](http://www.youtube.com/watch?v=Gqr3TCNNR2M "Groovestack CORE Jobs Demo")


spec/que/views/que_jobs_ext_spec.rb
```diff
@@ -0,0 +1,200 @@
...
+        end
+
+        it "should set status to complete when finished_at is present" do
+          locker
+          
+          sleep_until_equal(1) { DB[:que_lockers].count }
+          
+          TestJob.enqueue
+
+          sleep_until { locked_ids.count.positive? }
```
ZimbiX, line 143: I think there's a race condition here (and in the other tests) - in the case that this doesn't run until the job's already completed, the tests are going to get stuck / timeout. I know you've added a sleep to the job's run method (which is a separate issue), but it could still happen under high system load

spec/que/views/que_jobs_ext_spec.rb
```diff
@@ -0,0 +1,200 @@
...
+    class TestJob < Que::Job
+      include Que::JobMethods
+
+      def default_resolve_action
+        # prevents default deletion of complete jobs for testing purposes
+        finish
+      end
+
+      def run 
+        sleep(0.1)
```
ZimbiX, line 16: I'd like to avoid making the tests take longer from adding sleeps. How about using a mutex instead?

spec/spec_helper.rb
```diff
@@ -267,8 +271,16 @@ def internal_messages(event: nil)
     messages
   end
 
+  def pg_locks
```
ZimbiX, line 274: This is filtered, so I'd name it accordingly. `pg_advisory_locks`?

spec/spec_helper.rb
```diff
@@ -267,8 +271,16 @@ def internal_messages(event: nil)
...
+    DB[:pg_locks].where(locktype: 'advisory')
+  end
+
   def locked_ids
-    DB[:pg_locks].where(locktype: 'advisory').select_order_map(Sequel.lit("(classid::bigint << 32) + objid::bigint"))
+    pg_locks.select_order_map(Sequel.lit("(classid::bigint << 32) + objid::bigint"))
+  end
+
+  def locked_pids
+    pg_locks.select_order_map(Sequel.lit("pid"))
```
ZimbiX, line 283: I don't think `Sequel.lit` is necessary here

lib/que/utils/queue_management.rb
```diff
@@ -12,7 +12,10 @@ def clear!
       # Very old migrations may use Que.create! and Que.drop!, which just
       # created and dropped the initial version of the jobs table.
       def create!; migrate!(version: 1); end
-      def drop!;   migrate!(version: 0); end
+      def drop!
+        migrate!(version: 7) if Que.db_version >= 8
```
ZimbiX, line 16: Was this supposed to be committed?

spec/que/job.bulk_enqueue_spec.rb
```diff
@@ -44,6 +44,7 @@ def assert_enqueue(
       assert_equal expected_job_class.to_s, job[:job_class]
       assert_equal expected_args[i], job[:args]
       assert_equal expected_kwargs[i], job[:kwargs]
+      assert_equal job[:run_at], job[:first_run_at] if Que.db_version >= 8
```
ZimbiX, line 47: Why have you added all these db schema version condition checks? The specs only need to run against the latest db schema (other than the specs that test the migrations themselves)

spec/que/views/que_jobs_ext_spec.rb
```diff
@@ -0,0 +1,200 @@
...
+        assert_equal 0, jobs_ext_dataset.count
+      end
+
+      it "should include additional lock data" do
+        locker
+
+        sleep_until_equal(1) { DB[:que_lockers].count }
+
+        TestJob.enqueue
+
```
ZimbiX, line 70: Unnecessary extra blank line

docs/README.md
```diff
@@ -879,3 +880,15 @@ SET LOCAL que.skip_notify TO true;
 DELETE FROM que_jobs WHERE finished_at < (select now() - interval '7 days');
 COMMIT;
 ```
+
+## Que Jobs Ext View
```
ZimbiX, line 884: Markdown standard is to have a blank line between different elements/paragraphs - so here and after 'Additional Columns:'.

And downcase 'view' to match the headings capitalisation style.

ZimbiX (changes requested):
Thanks for the contribution!

Interesting idea. This does sound useful, but I'm not entirely sure the view needs to be in Que itself. I'll have a think, but I'm leaning towards adding it.

Could you squash the commits please?

hlascelles:
@maxschridde1494 We love to look of that dashboard. Can it be released as a gem or similar?

hlascelles:
Linking bit about "first run at" extra data: https://github.com/que-rb/que/pull/383#issuecomment-1768140767

maxschridde1494:
@ZimbiX Thanks for the feedback. I went ahead and committed the change requests and squashed the commits. Re race conditions in the specs, I removed the sleep call in the run method and refactored the specs. I believe this removed the race conditions in most (if not all) the specs. For the two specs that test the locker info getting pulled in and the running status getting correctly set when lock ids are present, I referenced the locker_spec.rb and implemented the pattern used there. Please let me know if you have any feedback for this implementation. Thanks!

maxschridde1494:
> @maxschridde1494 We love to look of that dashboard. Can it be released as a gem or similar?

Thanks for the positive feedback! We are currently working on packaging the new dashboard and will let you know as soon as it is available.

maxschridde1494:
> > @maxschridde1494 We love to look of that dashboard. Can it be released as a gem or similar?
> 
> Thanks for the positive feedback! We are currently working on packaging the new dashboard and will let you know as soon as it is available.

@hlascelles We are happy to announce the release of [create-groovestack](https://github.com/groovestack/create-groovestack). 

`create-groovestack` is a starter kit utility that generates a complete [Groovestack](https://talysto.com/tech/groovestack/) application that you can run immediately. Groovestack is a fullstack application framework consisting of [PostgreSQL](https://www.postgresql.org/), [Rails](https://rubyonrails.org/), [GraphQL](https://graphql.org/) and [React-Admin](https://marmelab.com/react-admin/). Here at [Talysto](https://talysto.com/), we have used Groovestack for many of our scalable applications across various industries.

`create-groovestack` includes several Groovestack CORE modules, including CORE Jobs, the module used in the demo above.

Issue #409, opened 2023-12-07
Proposal to change run_at default to clock_timestamp


hlascelles:
## Problem

Jobs enqueued in transactions can appear delayed as the default `run_at` is `now()`, which is a delayed time in a long transaction.

## Example

We know that the `que_jobs` table has this column: `run_at timestamp with time zone DEFAULT now() NOT NULL,`

We have some jobs that perform work, and enqueue another job. eg:

```ruby
class CalculateBilling < Que::Job
  def run(ref)
    ActiveRecord::Base.transaction do
      calculate_stuff(ref) # Takes 30 seconds
      SendBillingSms.enqueue(ref)
    end
  end
end
```

We then also have an alarm to check our jobs are running in a timely fashion by comparing when the job started with its `run_at`.

In the scenario above, this means the `SendBillingSms` job is enqueued with `run_at` as `now()`. But, `now()` does not mean "the time on the wall" in postgres. It means "the time the transaction started".

See: https://newcubator.com/blog/web/postgresql-timestamp-methods

_Both now() and clock_timestamp() are PostgreSQL functions. The main difference between them is that now() adds the time at the start of the transactions, while clock_timestamp() inserts the time at the execution of the statement._ 

This means all of our `SendBillingSms` jobs are instantaneously "30 seconds delayed", even if they are run the split second that they are visible to workers.

## Solution

We could/should change the column to have a default of `clock_timestamp()`.

ie: `run_at timestamp with time zone DEFAULT clock_timestamp() NOT NULL,`

This means that the intent of what the `run_at` of a job "to run now" should be is more accurate (probably, ie principle of least surprise). It also means jobs can be monitored for "delayed start" more easily and accurately.

Note, we do not want to use Ruby time in a `run_at` as it may be out, even by a second or two. We also would not like to pass a `run_at` in since:

1. We'd have to do it everywhere, or monkeypatch que
2. We'd have to do a DB lookup for what that time is, which is an extra call

Would you accept a PR for this? Thanks!

nickburgin:
That all seems well reasoned and perfectly reasonable to me. A PR would be welcome

hlascelles:
PR opened https://github.com/que-rb/que/pull/410

PR #410, opened 2023-12-11
Change run_at default to clock_timestamp()

hlascelles:
## Problem

Jobs enqueued in transactions can appear delayed as the default `run_at` is `now()`, which is a delayed time in a long transaction.

## Example

We know that the `que_jobs` table has this column: `run_at timestamp with time zone DEFAULT now() NOT NULL,`

We have some jobs that perform work, and enqueue another job. eg:

```ruby
class CalculateBilling < Que::Job
  def run(ref)
    ActiveRecord::Base.transaction do
      calculate_stuff(ref) # Takes 30 seconds
      SendBillingSms.enqueue(ref)
    end
  end
end
```

We then also have an alarm to check our jobs are running in a timely fashion by comparing when the job started with its `run_at`.

In the scenario above, this means the `SendBillingSms` job is enqueued with `run_at` as `now()`. But, `now()` does not mean "the time on the wall" in postgres. It means "the time the transaction started".

See: https://newcubator.com/blog/web/postgresql-timestamp-methods

_Both now() and clock_timestamp() are PostgreSQL functions. The main difference between them is that now() adds the time at the start of the transactions, while clock_timestamp() inserts the time at the execution of the statement._

This means all of our `SendBillingSms` jobs are instantaneously "30 seconds delayed", even if they are run the split second that they are visible to workers.

## Solution

We could/should change the column to have a default of `clock_timestamp()`.

ie: `run_at timestamp with time zone DEFAULT clock_timestamp() NOT NULL,`

This means that the intent of what the `run_at` of a job "to run now" should be is more accurate (probably, ie principle of least surprise). It also means jobs can be monitored for "delayed start" more easily and accurately.

Note, we do not want to use Ruby time in a `run_at` as it may be out, even by a second or two. We also would not like to pass a `run_at` in since:

1. We'd have to do it everywhere, or monkeypatch que
2. We'd have to do a DB lookup for what that time is, which is an extra call

Opened for issue: https://github.com/que-rb/que/issues/409

### DIFF
--- a/lib/que/migrations.rb
+++ b/lib/que/migrations.rb
@@ -4,7 +4,7 @@ module Que
   module Migrations
     # In order to ship a schema change, add the relevant up and down sql files
     # to the migrations directory, and bump the version here.
-    CURRENT_VERSION = 7
+    CURRENT_VERSION = 8
 
     class << self
       def migrate!(version:)
--- a/lib/que/migrations/8/down.sql
+++ b/lib/que/migrations/8/down.sql
@@ -1,0 +1,2 @@
+ALTER TABLE que_jobs
+ALTER COLUMN run_at SET DEFAULT now();
--- a/lib/que/migrations/8/up.sql
+++ b/lib/que/migrations/8/up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE que_jobs
+ALTER COLUMN run_at SET DEFAULT clock_timestamp();
--- a/spec/que/migrations.current_schema_spec.rb
+++ b/spec/que/migrations.current_schema_spec.rb
@@ -115,6 +115,19 @@ describe Que::Migrations, "current schema" do
     end
   end
 
+  describe "que_jobs table column defaults" do
+    it "should make sure that the default of run_at is clock_timestamp()" do
+      results = Que.execute (<<~SQL)
+        SELECT column_default
+        FROM information_schema.columns
+        WHERE table_name = 'que_jobs' AND column_name = 'run_at';
+      SQL
+      column_default = results.to_a.first.fetch(:column_default)
+
+      assert_equal column_default, "clock_timestamp()"
+    end
+  end
+
   describe "que_lockers table constraints" do
     let :locker_attrs do
       {
--- a/spec/que/migrations.state_trigger_spec.rb
+++ b/spec/que/migrations.state_trigger_spec.rb
@@ -43,7 +43,7 @@ describe Que::Migrations, "que_state trigger" do
 
     DB.transaction do
       result = yield
-      current_time = DB.get{now.function}
+      current_time = DB.get{clock_timestamp.function}
 
       id     ||= result
       run_at ||= current_time


hlascelles:
I don't know why the tests self-cancelled... I cannot rerun them?

ZimbiX:
I haven't assessed the changes, but the tests are cancelling because there's a failing test - it looks like they cancel because there's no point continuing the other test runs given there's a failure.

ZimbiX:
Ok, the failing test is expecting jobs to be processed via listen/notify with polling disabled, but none ever get locked - so it seems the change breaks listen/notify.

https://github.com/que-rb/que/blob/fb9278e728b6aca688b3ec8f5b340042a4942062/spec/que/locker_spec.rb#L266-L281

To run just that one test:

```bash
DATABASE_URL=postgres://que:que@localhost:5697/que-test TESTOPTS="-v --name='/locks only accepted jobs in a listen batch/'" USE_RAILS=true bundle exec rake spec
```

I suspected the `que_job_notify` trigger might be ignoring jobs where the run_at is in the future, and that does appear to be the case - the use of `now()` there is in conflict with `clock_timestamp()`:

https://github.com/que-rb/que/blob/fb9278e728b6aca688b3ec8f5b340042a4942062/lib/que/migrations/5/up.sql#L14-L22

ZimbiX:
After that, there are several other test failures due to asserting a run_at that's slightly off, due to an assumption in the tests that getting the time from the db (now with `clock_timestamp()`) multiple times within a transaction returns the same value.

hlascelles:
Thanks for the review/fail explanation!

We've been having some in depth discussions about this, and there are thoughts this may not even help with our current monitoring.

The simple example I gave will be pretty accurate (WAL stalls or other vagaries notwithtanding) but we frequently have jobs like this:

```ruby
class CalculateBilling < Que::Job
  def run(ref)
    ActiveRecord::Base.transaction do
      calculate_stuff(ref) # Takes 30 seconds

      SendBillingSms.enqueue(ref)

      work_out_things(ref) # Takes 10 seconds

      SendBillingEmails.enqueue(ref)
    end
  end
end
```

In this case `SendBillingSms` will always instantly be "delayed".

A more common example is the following. Here, even with this PR change, if we ran through 1000 accounts we will have some `SendBillingSms` as 1000 seconds delayed, and then a sliding line all the way up to "not delayed" as soon as the transaction commits.

```ruby
class CalculateAllBills < Que::Job
  def run(ref)
    ActiveRecord::Base.transaction do
      Accounts.all.each do |account|
        calculate_stuff(ref) # Takes 1 second
        SendBillingSms.enqueue(ref)
      end
    end
  end
end
```

We're debating if this PR change is slightly fixing half our cases, and making others misleading. Ideally we want a "at the end of this transaction, enqueue all jobs of any kind instantaneously". I guess it would be a more generic "auto" version of [bulk insert](https://github.com/que-rb/que/blob/master/docs/README.md#enqueueing-jobs-in-bulk).

Question to the floor: Does anyone have better ways of monitoring "delay by job class"?

ZimbiX:
I was indeed also wondering if this would fully fix your issue, as you've discovered, but I hadn't thought about it enough to articulate it.

Given you effectively want to record the committed at timestamp, I see what you mean about wanting to basically insert all the jobs at the end of the transaction. But is doing that yourself not a viable option? In your first example, you could put both enqueues at the end. And in your second example, you could first perform whatever slow calculations, then iterate over the results and enqueue all the jobs at once.

I understand it'd be great to not have to do that. A better way's not coming to me atm

hlascelles:
Yeah, the examples are very contrived and simplified. In reality we use [wisper](https://github.com/krisleech/wisper) a lot, so we sometimes "aren't aware" in a `run` method what other jobs may be enqueued elsewhere.

We're going to have more of a think on this and see. It may be that it is simply never possible for the run_at always to be "right" (whatever that may mean). Will have a look, and close this until then. :+1:

PR #417, opened 2024-02-12
Update README.md

ka8725:
The text was a bit hard to understand for me. I hope this version is better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,11 +90,9 @@ class ChargeCreditCard < Que::Job
       # Write any changes you'd like to the database.
       user.update charged_at: Time.now
 
-      # It's best to destroy the job in the same transaction as any other
-      # changes you make. Que will mark the job as destroyed for you after the
-      # run method if you don't do it yourself, but if your job writes to the DB
-      # but doesn't destroy the job in the same transaction, it's possible that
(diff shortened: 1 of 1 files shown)


ZimbiX (changes requested):
I appreciate what you're trying to do here, and I see how the current wording could be confusing; but I think this wording unfortunately introduces a different ambiguity.

By 'outside the transaction', it makes it sound like there will always be a transaction; but a transaction only exists within this example (and is a part of the usage this paragraph is supposed to be strongly recommending).

And we're losing mention of the important detail of when the job makes other writes to the DB. We need the transaction around the whole thing to ensure both the writes done by the job and the job deletion get committed together. Although on second thought, that is perhaps covered by the first sentence.

I instead suggest a very minor change to the original to break up the long sentence - replacing ', but ' with '; however':

```
# It's best to destroy the job in the same transaction as any other
# changes you make. Que will mark the job as destroyed for you after the
# run method if you don't do it yourself; however if your job writes to the DB
# but doesn't destroy the job in the same transaction, it's possible that
# the job could be repeated in the event of a crash.
```

ZimbiX (approved):
Thanks Andrei =)

PR #418, opened 2024-02-21
update readme with link to Que::View

kortirso:
I'd like to introduce Que::View Rails engine for using instead of Que::Web for UI

With this engine the number of dependencies is reduced compared to Que::Web, no more Sinatra, jQuery, some redundant styles and js, and looks like Que::Web was updated 2 years ago last time

Que::View has additional features:
- processes metrics
- queue metrics

@nickburgin @ZimbiX @siegy22 what do you think?


ZimbiX (approved):
Cool! I'll have to check it out

> the number of dependencies is reduced

Given using Rails over Sinatra increases the overall number/size of dependencies, I think you mean the number of non-Rails dependencies =P

PR #419, opened 2024-03-01
[BOX-840] Opslevel update

rspesia1:
Added the opslevel.yml to achieve bronze level in OpsLevel.

### DIFF
--- a/opslevel.yml
+++ b/opslevel.yml
@@ -1,0 +1,9 @@
+service:
+  name: que
+  owner: back_office_experience_box
+  description: High-performance job queue.
+  language: Ruby
+  tier: tier_1
+  lifecycle: pre-release
+  tags:
+  - key: type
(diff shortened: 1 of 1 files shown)


opslevel.yml
```diff
@@ -0,0 +1,9 @@
+service:
+  name: que
+  owner: back_office_experience_box
+  description: High-performance job queue.
+  language: Ruby
+  tier: tier_1
+  lifecycle: pre-release
+  tags:
+  - key: type
```
orlandoCamarillo, line 9: the value of this key is missing

orlandoCamarillo: seems like you forgot to add the type value at the very end of the file
`value: <the type of this project>`
for example
  ```
- key: type
    value: service
```

orlandoCamarillo (approved):

PR #420, opened 2024-03-04
Support bulk enqueue with differing class and options 

dtcristo:
Follow-up to https://github.com/que-rb/que/pull/340.

Allow bulk enqueue of multiple different job classes and differing job options in a single `.bulk_enqueue` block. Each job can now differ by job class, queue, priority, run at and tags (in addition to args and kwargs).


dtcristo:
I rebased this PR on top of https://github.com/que-rb/que/pull/421 as CI was broken due to Bundler version.

lib/que/job.rb
```diff
@@ -29,21 +29,18 @@ class Job
...
         (queue, priority, run_at, job_class, args, kwargs, data, job_schema_version)
         SELECT
-          coalesce($1, 'default')::text,
-          coalesce($2, 100)::smallint,
-          coalesce($3, now())::timestamptz,
-          $4::text,
-          args_and_kwargs.args,
-          args_and_kwargs.kwargs,
-          coalesce($6, '{}')::jsonb,
+          coalesce(queue, 'default')::text,
```
ZimbiX, line 35: I didn't know you could refer to these by name! Excellent

lib/que/job.rb
```diff
@@ -167,49 +176,43 @@ def _bulk_enqueue_insert(args_and_kwargs_array, job_options: {}, notify:)
...
-          values_array.map(&method(:new))
-        end
+        end.compact
+
+        values_array =
+          Que.transaction do
+            Que.execute('SET LOCAL que.skip_notify TO true') unless notify
+            Que.execute(
+              :bulk_insert_jobs,
+              [Que.serialize_json(jobs_attrs.map { |attrs| attrs.except(:klass) })]
```
ZimbiX, line 212: Ruby 2.7 annoyingly does not have `Hash#except`. iirc, I've used `Hash#reject` for this previously.

spec/que/job.bulk_enqueue_spec.rb
```diff
@@ -4,14 +4,14 @@
 
 describe Que::Job, '.bulk_enqueue' do
   def assert_enqueue(
-    expected_queue: 'default',
-    expected_priority: 100,
-    expected_run_at: Time.now,
-    expected_job_class: Que::Job,
-    expected_result_class: nil,
+    expected_queues: ['default'].cycle(10).to_a,
```
ZimbiX, line 7: How about just `[value] * 10` for these?

spec/que/job.bulk_enqueue_spec.rb
```diff
@@ -278,21 +296,13 @@ def assert_enqueue(
     end
   end
 
-  it "should raise when job_options are passed to .enqueue rather than .bulk_enqueue" do
-    assert_raises_with_message(Que::Error, "When using .bulk_enqueue, job_options must be passed to that method rather than .enqueue") do
```
ZimbiX, line 282: We should also update the [bulk enqueue documentation](https://github.com/que-rb/que/tree/master/docs#enqueueing-jobs-in-bulk) to remove the applicable limitations

ZimbiX (changes requested):
Great work! Thanks for this :D

lib/que/job.rb
dtcristo: Yeah, I noticed this broke CI for Ruby 2.7. I'll fix this.

spec/que/job.bulk_enqueue_spec.rb
dtcristo: Will do.

lib/que/job.rb
dtcristo: Yeah, this is only possible because it's using [json_populate_recordset](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-PROCESSING-TABLE). A nice cleanup IMO.

PR #421, opened 2024-03-05
Fix bundler in CI

dtcristo:
[Bundler 2.5 dropped support for Ruby 2.7](https://github.com/rubygems/rubygems/releases/tag/bundler-v2.5.0). Since we are testing with Ruby 2.7, we should pin the version of Bundler to 2.4.x since this is the last version that's compatible with 2.7.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
       run: |
         sudo apt-get -yqq install libpq-dev postgresql-client
         createdb que-test
-        gem install bundler
+        gem install bundler --version '~> 2.4.22'
         bundle install --jobs 4 --retry 3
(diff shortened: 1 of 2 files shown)


ZimbiX:
Thanks. [This](https://github.com/que-rb/que/blob/fb9278e728b6aca688b3ec8f5b340042a4942062/Dockerfile#L8) should also be changed to keep the version in sync

dtcristo:
> Thanks. [This](https://github.com/que-rb/que/blob/fb9278e728b6aca688b3ec8f5b340042a4942062/Dockerfile#L8) should also be changed to keep the Bundler version in sync

Done. Should be ready to go now.

owst:
@ZimbiX this seems ready to go - can it be merged?

PR #426, opened 2024-04-29
Fix server? method

donbobka:
My understanding is that the `server?` method is supposed to return `true` if the current process is the que server. It looks like the method is missing a `!`. Right now, it is returning `false` for the queue server and `true` for non-queue processes.

### DIFF
--- a/lib/que.rb
+++ b/lib/que.rb
@@ -80,7 +80,7 @@ module Que
     end
 
     def server?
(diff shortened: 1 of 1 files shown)


ZimbiX:
Thanks for picking this up!

Introduced in https://github.com/que-rb/que/pull/382, CC @mnbbrown

PR #428, opened 2024-07-24
Log full job details rather than just job_id

owst:
A partial revert of https://github.com/que-rb/que/commit/c43eb24aebcc5054c299cfed00690fce9add36ac which didn't include any reasoning/justification for the log simplification.

It's useful to be able to log details of the job, particularly the job_class and arguments; if this additional detail isn't desired then it can be removed by using a custom `Que.log_formatter`, for example:

```ruby
Que.log_formatter = proc do |data|
  if (job = data.delete(:job))
    data[:job_id] = job[:id]
  end

  JSON.dump(data)
end
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -349,7 +349,7 @@ que --worker-count 1
 By default, Que logs important information in JSON to either Rails' logger (when running in a Rails web process) or STDOUT (when running via the `que` executable). So, your logs will look something like:
 
 ```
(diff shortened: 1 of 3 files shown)


paul110 (approved):

ZimbiX (approved):
From memory, at GreenSync we instead relied on instrumentation from Datadog APM (implemented via Que middleware). But more detail in the logs - particularly the job class and arguments - sounds like a good idea to me. I vaguely recall being annoyed about the omission of that detail way back when I started using Que. Shame about the lack of context for its removal.

I would hope that the extra log volume this adds would not need to made configurable; but that could be added if needed.

I think the chances are low that the removal of `job_id` from the top-level logged hash would cause any issues for people depending on it.

Issue #282, opened 2020-03-26
Don't require relative entrypoint file path to begin with ./


ZimbiX:
When specifying the entrypoint file path without a `./`, it doesn't work:

```
➜ bundle exec que config/jobs_environment.rb  
Could not load file 'config/jobs_environment.rb'
```

vs:

```
➜ bundle exec que ./config/jobs_environment.rb
Que waiting for jobs...
```

This is confusing, and it took us digging through the source code to figure out that `./` was required. We have filed a related issue for revealing the true error when loading this file: https://github.com/que-rb/que/issues/280

PR #433, opened 2024-11-23
Add more descriptive error to file loading

TheBlackArroVV:
After trying to run it locally I found out, that an original error about the unavailability to run a file is not descriptive. So, I decided to add an error to make it more obvious.

### DIFF
--- a/lib/que/command_line_interface.rb
+++ b/lib/que/command_line_interface.rb
@@ -196,13 +196,7 @@ OUTPUT
         end
 
         args.each do |file|
-          begin
-            require file
-          rescue LoadError => e
-            output.puts "Could not load file '#{file}': #{e}"
-            output.puts "Your file path should look like you're running a script, e.g. `./path/to/file.rb`" unless file.start_with?('./')
-            return 1
-          end
+            require File.expand_path(file)
         end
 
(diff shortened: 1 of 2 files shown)


ZimbiX:
Resolves https://github.com/que-rb/que/issues/280 and https://github.com/que-rb/que/issues/282.

imo the `begin` block should be removed so that errors from the require path or required code can be shown normally. And perhaps requiring using the full path to the given file.

TheBlackArroVV:
Hi, @ZimbiX,
Can you elaborate a bit about wdym? If you you mean just rendering a standard LoadError I think it's not gonna help with a problem I had, it would still be unclear how it should be used

ZimbiX:
Something like:

```ruby
args.each do |file|
  require whatever_method_returns_a_full_file_path(file)
end
```

That way, all forms of relative file path can be accepted. Then if an exception is raised due to the given file not existing, that should be pretty easy to understand already, I'd think.

Edit: Looks like `File.expand_path` or `File.realpath` does the trick

TheBlackArroVV:
@ZimbiX, got it, makes sense
Can you check now?

TheBlackArroVV:
Hey, @ZimbiX,
Cam you take a look?

PR #440, opened 2025-11-14
Don't log non-error "undefined method 'fields' for nil".

bglimepoint:
`result` is nil when running `DELETE FROM public.que_jobs WHERE id = $1::bigint`, which results in an error log.

---

Not sure whether this is something specific to us, but we're running 2.4.1 and are often seeing `"error":"NoMethodError: undefined method 'fields' for nil"` in our Que logs. Looks like this is where it's coming from (inside `convert_result` when it calls `result.fields`).


lib/que/connection.rb
```diff
@@ -75,7 +75,7 @@ def execute(command, params = [])
         }
       end
 
-      convert_result(result)
+      convert_result(result) unless result.nil?
```
oeoeaio: I note that there are other call sites for `convert_result`, I'm wondering if this is better placed as a guard clause inside `convert_result` rather than out here?

PR #444, opened 2025-12-31
Fix broken test for Test: Ruby 3.0, Rails 6.0 + 6.1, PostgreSQL 14

lee-treehouse:
Failing test https://github.com/que-rb/que/actions/runs/20613438072/job/59201910077#step:5:131 for `Test: Ruby 3.0, Rails 6.0, PostgreSQL 14` seems to be explained [here](https://stackoverflow.com/a/79361034)

### DIFF
--- a/spec/gemfiles/Gemfile-rails-6.0
+++ b/spec/gemfiles/Gemfile-rails-6.0
@@ -5,6 +5,7 @@ gem 'que', path: '../..'
 group :development, :test do
   gem 'rake'
 
+  gem 'concurrent-ruby', '1.3.4'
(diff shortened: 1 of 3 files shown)


ZimbiX:
Thanks for looking into this!

With those Gemfiles, I like that the only difference between them is the version of the Rails gems, and I'd like to keep it that way; so I think I'd actually prefer the other solution posted in that StackOverflow - adding `require 'logger'` to the spec_helper after these lines:

https://github.com/que-rb/que/blob/fbd57fda1fe148cda0fa527d49559dbe0033fca6/spec/spec_helper.rb#L11-L13

ZimbiX (approved):
Cheers!

PR #449, opened 2026-09-03
Ruby 3.4 support

salimepoint:
Adds Ruby 3.3 and 3.4 to CI and fixes the three things that stood in the way.

1. mutex_m removal broke suite loading

Ruby 3.4 removed mutex_m from the standard library. The pinned minitest ~> 5.10.1 requires it, so the suite failed before a single test ran. minitest ~> 5.25.0 has no such dependency and still supports Ruby 2.7, so one pin covers the whole existing matrix — bumped in the root Gemfile and in all five spec/gemfiles/Gemfile-rails-*.

2. Hash#inspect changed its rendering

Ruby 3.4 renders symbol keys as {key: 1} rather than {:key=>1}. listener_spec hardcoded the old rendering of a hash that Que::Listener inspects into an error message. Deriving the expectation with #inspect instead of writing it out makes it hold on every Ruby version.

3. Temp spec file names collided under the newer minitest (9f10965)

This one is a prerequisite for the bump rather than a Ruby 3.4 issue, so it's a separate commit.

minitest 5.16 added srand Minitest.seed to Test.runnable_methods, which runs once per suite — so from that version on, the global PRNG replays the same sequence at the start of every describe block. command_line_interface_spec's write_file built its file name from Digest::MD5.hexdigest(rand.to_s), so specs in different describe blocks were handed the same name. The second require of that path is a no-op — it's already in $LOADED_FEATURES — so LOADED_FILES never gets an entry for it and the assertions on it fail.

Which specs collide moves around with the seed, so the failures read as flakes rather than as a bug. Under minitest 5.25, every one of six seeds tried produced between one and three of them. Switching to SecureRandom.hex(16) fixes it, as SecureRandom is unaffected by srand.

CI

Adds { ruby: 3.3, rails: 7.2, postgres: 14 } and { ruby: 3.4, rails: 7.2, postgres: 14 }. Rails 7.2 only — the older Rails gemfiles are already excluded from the newer Rubies they don't support.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,6 +18,8 @@ jobs:
         - { ruby_version: '3.2', rails_gemfile: '7.2', postgres_version: '12' }
         - { ruby_version: '3.2', rails_gemfile: '7.2', postgres_version: '13' }
         - { ruby_version: '3.2', rails_gemfile: '7.2', postgres_version: '14' }
+        - { ruby_version: '3.3', rails_gemfile: '7.2', postgres_version: '14' }
+        - { ruby_version: '3.4', rails_gemfile: '7.2', postgres_version: '14' }
         exclude: # Rails 7.2 is not compatible with Ruby < 3.1
         - ruby_version: '2.7'
           rails_gemfile: '7.2'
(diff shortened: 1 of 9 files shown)


oeoeaio (approved):
Thanks for the contribution @salimepoint!

This looks good. Will review the test matrix once I've had a look at some other PRs.

